PR #2, opened 2017-02-27
Update ImageData IDL to match the new color space settings of Canvas

zakerinasab:


CanvasColorSpaceProposal.md
```diff
@@ -140,18 +140,23 @@ IDL
 <pre>
 typedef (Uint8ClampedArray or Uint16ClampedArray or Float32Array) ImageDataArray;
 
-[Constructor(unsigned long sw, unsigned long sh, optional CanvasColorSpace colorSpace = "legacy-srgb"),
- Constructor(ImageDataArray data, unsigned long sw, optional unsigned long sh, optional CanvasColorSpace colorSpace),
+dictionary ImageDataColorSettings {
+  CanvasColorSpace colorSpace = "srgb";
+  CanvasPixelFormat pixelFormat = "8-8-8-8";
```
junov: Hmmm, I am not sure what is the use of pixel format.  Isn't it implied in the type of the ImageData (ImageData vs Float32ImageData)?

CanvasColorSpaceProposal.md
```diff
@@ -140,18 +140,23 @@ IDL
...
   readonly attribute unsigned long width;
   readonly attribute unsigned long height;
   readonly attribute ImageDataArray data;
-  readonly attribute CanvasColorSpace colorSpace;
+  readonly attribute ImageDataColorSetting imageDataColorSettings;
 };
 </pre>
 
-* getImageData() produces an ImageData object in the same color space as the source canvas, using an ImageDataArray of a type that is appropriate for the specified pixelFormat (smallest possible numeric size that guarantees no loss of precision)
+* getImageData() produces an ImageData object with the same color settings (colorSpace and pixelFormat) as the source canvas, using an ImageDataArray of a type that is appropriate for the specified pixelFormat (smallest possible numeric size that guarantees no loss of precision).
```
junov: ImageData can only by 8-8-8-8

CanvasColorSpaceProposal.md
junov: The types possible for ImageData do not map 1 to 1 with CanvasPixelFormat.  I think a new enum is necessary

junov: "uint8", "uint16", or "float32"

zakerinasab: I think we need to keep ImageDataColorSettings similar to CanvasRenderingContext2DSettings because PutImageData needs to know the pixel format of the image data it wants to color convert to the color space of the destination canvas. If we create an ImageData with, for example, p3 and Uint16ClampedArray, how PutImageData knows if the p3 is in 10-10-10-2 pixel format or 12-12-12-12 pixel format? I also believe we should clarify which type of ImageDataArray must be used for 10-10-10-2 and 12-12-12-12 pixel formats.

zakerinasab: Also what happens if the constructor is called with an ImageDataArray of type Float32Array but without specifying the optional color settings? Shall we have a "default" color space for Float32Array, like p3 or rec2020?

junov: > how PutImageData knows if the p3 is in 10-10-10-2 pixel format or 12-12-12-12 pixel format?

The ImageData is in 16-16-16-16 format, which is implied in the Uint16ClampedArray. putImageData will convert pixels to what ever format the canvas is in (e.g. 10-10-10-2 or 12-12-12-12)

CanvasColorSpaceProposal.md
```diff
@@ -138,21 +138,35 @@ ImageBitmap objects are augmented to have an internal color space attribute of t
 
 IDL
 <pre>
+
+enum ImageDataStorageFormat {
```
junov: "type" instead of "format". The word format implies more than just type. e.g. the number of channels and/or their order.

CanvasColorSpaceProposal.md
```diff
@@ -138,21 +138,35 @@ ImageBitmap objects are augmented to have an internal color space attribute of t
...
   readonly attribute unsigned long width;
   readonly attribute unsigned long height;
   readonly attribute ImageDataArray data;
-  readonly attribute CanvasColorSpace colorSpace;
+  readonly attribute ImageDataColorSettings colorAttributes;
 };
 </pre>
 
-* getImageData() produces an ImageData object in the same color space as the source canvas, using an ImageDataArray of a type that is appropriate for the specified pixelFormat (smallest possible numeric size that guarantees no loss of precision)
+* When using the constructor with ImageDataArray parameter, the array type must respect ImageDataColorSettings.ImageDataStorageFormat. Otherwise, the return object will be undefined.
```
junov: Would prefer: When using the constructor that takes an ImageDataArray parameter, the "storageFormat" setting is ignored.

CanvasColorSpaceProposal.md
```diff
@@ -138,21 +138,35 @@ ImageBitmap objects are augmented to have an internal color space attribute of t
...
+* getImageData() produces an ImageData object with the same color space as the source canvas, using an ImageDataArray of a type that is appropriate for the pixelFormat of the source canvas (smallest possible numeric size that guarantees no loss of precision).
 * putImageData() performs a color space conversion to the color space of the destination canvas.
+* The proper ImageDataStorageFormat for "10-10-10-2" and "12-12-12-12" canvas pixel formats is "uint16". getImageData() and putImageData() use bit shifting for numerical scaling.
```
junov: Actually, the correct conversion Is more complex that just shifting if you want to get the closest value in the new space. For example 10 to 16 bit conversion is: (x << 6) + (x >> 4).  Besides, there is no need to document the conversion algorithm here.

PR #22, opened 2017-10-10
Add linear-srgb and extended-srgb

ccameron-chromium:
Add linear-srgb and extended-srgb color spaces, and remove p3 and bt2020.
Add notes on extended pixel interpretation.
Remove support for 10-10-10-2.

PR #23, opened 2017-11-07
Add support for ITU Rec 2100 PQ narrow and full quantization

palemieux:
The objective is to add support for use cases when presentation of the canvas requires a greater dynamic range than that provided by the sRGB color space, and when that presentation requires an absolute luminance value.

The approach to converting sRGB colors to Rec 2100 PQ narrow and full quantization colors is the same used in [TTML2](https://www.w3.org/TR/ttml2/#style-attribute-luminanceGain)

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -1,7 +1,7 @@
 # Color managing canvas contents
 
 ## Use Case Description
-* Contents displayed through a canvas element should be color managed in order to minimize differences in appearance across browsers and display devices. Improving color fidelity matters a lot for artistic uses (e.g. photo and paint apps) and for e-commerce (product presentation).
+* Contents displayed through a canvas element should be color managed in order to minimize differences in appearance across browsers and display devices. Improving color fidelity matters a lot for artistic uses (e.g. photo and paint apps), for e-commerce (product presentation) and when canvas element are presented alongside video content.
 * Canvases should be able to take advantage of the full color gamut and dynamic range of the display device.
 * Creative apps that do image manipulation generally prefer compositing, filtering and interpolation calculations to be performed in a linear color space.
 
@@ -44,6 +44,8 @@ IDL:
 enum CanvasColorSpace {
   "srgb", // default
   "rec2020",
+  "rec2100pq",
+  "rec2100pqn",
   "p3",
 };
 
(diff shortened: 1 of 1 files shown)


junov:
Thanks for proposing this modification.  As you can see in PR #22 we are trying to reduce the surface of this API by limiting the number of natively supported color spaces to the bare minimum. Instead of creating new colorspaces to support HDR standards, could we consolidate the PQ behavior into existing configurations?  @ccameron-chromium WDYT?

palemieux:
@junov Rec 2100 PQ specifies colors with absolute luminance (in cd/m^2), instead of normalized luminance like sRGB. It is not clear how absolute luminance colors would be mapped to/from `extended-srgb`.

CanvasColorSpaceProposal.md
```diff
@@ -87,6 +89,21 @@ canvas.getContext('2d', { colorSpace: "p3", pixelFormat: "10-10-10-2", linearPix
 * toDataURL/toBlob convert image data to the rec-2020 color space.
 * Images with no color profile, when drawn to the canvas, are assumed to be in the sRGB color space, and are converted to rec2020 for the purpose of the draw.
 
+#### The "rec2100pq" color space
+* Support is optional.
+* User agents that select the rec2100pq color gamut in CSS media queries must support this color space.
```
svgeesus: There is a Media Query called [`rec2020`](https://drafts.csswg.org/mediaqueries-4/#color-gamut) but there is no Meda Query rec2100pq or rec2100pqn.

Do you mean that only ultrawide gamut (and HDR capable) devices must support it?

For example, if a device reports true for `p3` (so it is wide gamut) and false for `rec2020` (so it is not ultrawide gamut) it should still support content delivered in rec2020 colorspace, no? I mean, that is basically what all 4k HDR TVs do now.

CanvasColorSpaceProposal.md
```diff
@@ -105,11 +122,15 @@ The pixelFormat attributes specifies the numeric types to be used for storing pi
...
+    ( window.matchMedia("(color-gamut: p3)").matches ? "p3" : 
+      ( window.matchMedia("(color-gamut: rec2100pq)").matches ? "rec2100pq" :
+        ( window.matchMedia("(color-gamut: rec2100pqn)").matches ? "rec2100pqn" : "srgb" )
+      )
+    );
 </pre>
 
 #### Selecting the best pixelFormat for the user agent's display device
-Selection should be based on the best color space match (see above). For srgb, at least 8 bits per component is recommended; for p3, 10 bits; and for rec2020, 12 bits.  The float16 format is suitable for any colorspace.  There may soon be a proposal to add a way of detecting HDR displays, possibly something like "window.screen.isHDR()" (TBD), which would be a good hint to use the float16 format.
+Selection should be based on the best color space match (see above). For srgb, at least 8 bits per component is recommended; for p3, 10 bits; and for rec2020 and rec2100pq and rec2100pqn, 12 bits.  The float16 format is suitable for any colorspace.  There may soon be a proposal to add a way of detecting HDR displays, possibly something like "window.screen.isHDR()" (TBD), which would be a good hint to use the float16 format.
```
svgeesus: Media Queries 5 will allow querying both the screen and the video plane (in the cases where that is different, such as TVs) for color gamut and for dynamic range. See https://github.com/w3c/csswg-drafts/issues/4471#issuecomment-549178832

CanvasColorSpaceProposal.md
```diff
@@ -171,6 +193,7 @@ interface ImageData {
 * putImageData() performs a color space conversion to the color space of the destination canvas.
 * Data returned by getImageData() or passed to putImageData() are assumed to be in linear color space.
 * If the image data color space is not "srgb" or the storage format is not "uint8", the data passed to putImageData() is assumed to be in linear color space.
+* the `luminanceGain` property is used to map sRGB colors to colors with greater dynamic range, including those in the `rec2100pq` and `rec2100pqn` colorspaces, as specified at (https://www.w3.org/TR/ttml2/#style-attribute-luminanceGain).
```
svgeesus: Normalizing sRGB to 1 (initial value of luminanceGain) times 80 = 80 cd/m^2 is spec compliant but suboptimal, and does not reflect current practice. A luminanceGain of 2 would seem more appropriate.

svgeesus:
@palemieux is this PR still current, or have more recent events overtaken it?

palemieux:
@svgeesus Wow... I had totally forgotten about this. I think it is still germane since (a) canvas does not handle HDR (yet) and (b) the topic is under discussion within the Color CG.

palemieux:
Overtaken by #60

PR #28, opened 2020-04-10
Split out attribs into [color-space, encoding, linearity]

jdashg:
There's renewed interest in this from the WebGL WG, so I'm taking a shot at moving this along.

Much of this is advised by some thought I've been putting into our story for sRGB and WebGL: https://hackmd.io/@jgilbert/sRGB-WebGL

A bunch of changes here, so there'll be a lot to talk about.
* Splitting out attribs into [color-space, encoding, linearity] adds specificity at some cost of verbosity and an increase of jargon.
* Remove linear=physical association, which is a common source of confusion (and contention between APIs).
* Decouple radiance-linearity from rec2020, and make rec2020 match the CSS color spec. (Therefore it no longer uses sRGB primaries)
* Added clarifying language about encoding-vs-value and operations

I wouldn't call this done, but we'd love to get this moving again.

Cc: @kenrussell, @grorg, @RafaelCintron, @MarkCallow

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -7,10 +7,10 @@
 ### Current Limitations
 * The color space of canvases is undefined in the current specification, though de facto sRGB.
 * The bit-depth of canvases is currently fixed to 8 bits per component, which is below the capabilities of some monitors. Monitors with higher contrast ratios require more bits per component to avoid banding.
-* Color encoding and blending is de facto luminance-linear unorm8, which allocates too little precision on low values (and too much precision on high values), leading to poor handling of dark scenes and content.
+* Color encoding and blending is de facto perceptually-linear unorm8, which allocates too little precision on low values (and too much precision on high values), leading to poor handling of dark scenes and content.
 
 ### Current Usage and Workarounds
-The lack of color space interoperability is hard to work around. With some browser implementations that color correct images drawn to canvases by applying the display profile, apps that want to use canvases for color corrected image processing are stuck doing convoluted workarounds, such as:
+The lack of color space interoperability is hard to work around. For some browser implementations which color-correct images drawn to canvases by applying the display profile, apps that want to use canvases for color corrected image processing are stuck doing convoluted workarounds, such as:
 * reverse-engineer the display profile by drawing test pattern images to the canvas and inspecting the color corrected result via getImageData
 * bypass CanvasRenderingContext2D.drawImage() and use image decoders implemented in JavaScript to extract raw image data that was not tainted by the browser's color correction behavior.''
 
@@ -26,7 +26,20 @@ Some implementations convert images drawn to canvases to the sRGB color space. T
 
 * <cite>[https://github.com/whatwg/html/issues/299]</cite> <blockquote><p>Allow 2dcontexts to use deeper color buffers</p></blockquote>
 * <cite>[https://bugs.chromium.org/p/chromium/issues/detail?id=425935]</cite> <blockquote><p>Wrong color profile with 2D canvas</p></blockquote>
-* Engineers from the Google Photos, Maps and Sheets teams have expressed a desire for canvases to become color managed.  Particularly for the use case of resizing an imaging, using a canvas, prior to uploading it to the server, to save bandwidth. The problem is that the images retrieved from a canvas are in an undefined color space and no color space information is encoded by toDataURL or toBlob.
+* Engineers from the Google Photos, Maps and Sheets teams have expressed a desire for canvases to become color managed.  Particularly for the use case of resizing an image, using a canvas, prior to uploading it to the server, to save bandwidth. The problem is that the images retrieved from a canvas are in an undefined color space and no color space information is encoded by toDataURL or toBlob.
+
+## Background
+
+Color spaces are generally tuples of:
+* Chromaticity coordinates (X,Y in CIE 1931 space)
+  * Primaries (coords for 1.0 for each of R/G/B)
+  * White-point (coord for white)
+* Transfer function (often "gamma")
+
+Within the same chromaticity, 0.0 will always be min-brightness and 1.0 is max-brightness.
+However, because of the transfer function, 0.5 will not be the average physical brightness of 0.0 and 1.0.
+You can see this by making stripes of 1.0 and 0.0, and matching that pattern's brightness to a solid grey background.
+(On my machine, striped 0/1/0/1 appears equally bright as #ACACAC, or 0.67)
 
 ## Proposed Solution
 
@@ -49,7 +62,7 @@ enum CanvasColorSpaceEnum {
 enum CanvasColorEncodingEnum {
   "unorm8",      // default, 0.5 encoded as 0x80
   "unorm8-srgb", // 0.5 encoded as 0xbc
-  "float16",
+  "float16",     // IEEE 754
 };
 
 // Feature detection:
@@ -122,7 +135,7 @@ The colorEncoding attributes specifies the encoding to be used for storing pixel
 * Support for "unorm8" is mandatory. All other encodings are optional.
 * When an unsupported encoding is requested, the encoding shall fall back to "unorm8".
 * The alpha channel is always interpreted as if clamped to [0,1].
-* Float RGB channel values outside of [0,1] range can be used to represent colors outside of the chosen color gamut. This allows float pixel formats to represent all possible colors and brightness levels. How values outside of [0,1] are displayed depends on the capabilites of the device and output display. Some implementations may simply clamp these values to [0,1]. If the device and display are capable, a (luminance-linear) pixel value of (2,2,2) should be twice as bright as (1,1,1).
+* Float RGB channel values outside of [0,1] range can be used to represent colors outside of the chosen color gamut. This allows float pixel formats to represent all possible colors and brightness levels. How values outside of [0,1] are displayed depends on the capabilites of the device and output display. Some implementations may simply clamp these values to [0,1]. If the device and display are capable, a (perceptually-linear) pixel value of (2,2,2) should be twice as bright as (1,1,1).
 * Operations on encoded values always operate on decoded values, not the encoded bits.
     * I.e. in "unorm8-srgb" encoding, 0xff (1.0) minus 0xbc (0.5) equals 0xbc (0.5).
 
@@ -199,8 +212,7 @@ Authors of games and imaging apps are expected to be enthusiastic adopters.
 
 * Should float16 allow alpha values outside [0,1] range at any stage of the pipeline? What would they mean?
 
-* Should there be API-level support for radiance-linear (non-perceptual) color spaces or blending?
-    * This might be useful for canvas2d, but WebGL can already support most of this use case.
+* Should there be API-level support for mixing chromaticities and transfer functions, including use of no-op transfer functions?
 
 * Should it be "rgba8" and "rgba16f" instead of "unorm8" and "float16"?
 
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -6,8 +6,9 @@
 * Some applications prefer that compositing, filtering and interpolation calculations be performed in a physically based linear color space.
 
 ### Current Limitations
-* The color space of canvases is undefined in the current specification.
+* The color space of canvases is undefined in the current specification, though de facto sRGB.
 * The bit-depth of canvases is currently fixed to 8 bits per component, which is below the capabilities of some monitors. Monitors with higher contrast ratios require more bits per component to avoid banding.
+* Color encoding and blending is de facto luminance-linear unorm8, which allocates too little precision on low values (and too much precision on high values), leading to poor handling of dark scenes and content.
 
 ### Current Usage and Workarounds
 The lack of color space interoperability is hard to work around. With some browser implementations that color correct images drawn to canvases by applying the display profile, apps that want to use canvases for color corrected image processing are stuck doing convoluted workarounds, such as:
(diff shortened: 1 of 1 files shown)


CanvasColorSpaceProposal.md
```diff
@@ -63,56 +104,72 @@ partial interface CanvasRenderingContext2D {
...
+The colorEncoding attributes specifies the encoding to be used for storing pixel channel color values.
+* Support for "unorm8" is mandatory. All other encodings are optional.
+* When an unsupported encoding is requested, the encoding shall fall back to "unorm8".
+* The alpha channel is always interpreted as if clamped to [0,1].
+* Float RGB channel values outside of [0,1] range can be used to represent colors outside of the chosen color gamut. This allows float pixel formats to represent all possible colors and brightness levels. How values outside of [0,1] are displayed depends on the capabilites of the device and output display. Some implementations may simply clamp these values to [0,1]. If the device and display are capable, a luminance-linear pixel value of (2,2,2) should be twice as bright as (1,1,1).
+* Operations on encoded values always operate on decoded values, not the encoded bits.
+    * I.e. in "unorm8-srgb" encoding, 0xff (1.0) minus 0xbc (0.5) equals 0xbc (0.5).
+
+
+#### The colorLinearity context creation attribute
```
lexaknyazev, line 146: The spec should probably give more information on how this field is expected to be utilized/implemented.

CanvasColorSpaceProposal.md
```diff
@@ -63,56 +104,72 @@ partial interface CanvasRenderingContext2D {
...
-* When an unsupported format is requested, the format shall fall back to "8-8-8-8".
-* With formats that use integer numeric types for the color channels, the canvas pixel buffer shall store non-linear color values, encoded using the transfer curves prescribed by the specification of the current color space.
-* With formats that use floating-point numeric types for the color channels, the canvas pixel buffer shall store linear (i.e unencoded) color values.
-* Float values outside of [0,1] range can be used to represent colors outside of the chosen color gamut. This allows float pixel formats to represent all possible colors and brightness levels. How values outside of [0,1] are displayed depends on the capabilites of the device and output display. Some implementations may simply clamp these values to [0,1]. If the device and display are capable, a pixel value of (2,2,2) should be twice as bright as (1,1,1).
+* toDataURL/toBlob produce resources tagged as being in the sRGB color space, if the encoding format supports colorspace tagging or embedded color profiles.
+    * Exception: User agents that implement the legacy behavior must not encode any color space metadata.
```
lexaknyazev, line 128: PNG is specified to use sRGB primaries/transfer by default (when no extra metadata is present).

CanvasColorSpaceProposal.md
```diff
@@ -26,34 +27,74 @@ Some implementations convert images drawn to canvases to the sRGB color space. T
...
-
 IDL:
 <pre>
-enum CanvasColorSpace {
+// Feature enums:
+
+enum CanvasColorSpaceEnum {
   "srgb", // default
-  "linear-srgb",
+  "rec-2020",
```
kenrussell, line 48: @ccameron-chromium and @FredrikHubinette have worked extensively on Chrome's color profile support and would ideally comment on this change. There were some thoughts that "extended-srgb" would be a more portable choice for a color space supporting wide-color gamut and HDR use cases. Agree that the previous iteration of this spec seemed to have accidentally dropped the rec-2020 enum without fully removing references to it.

CanvasColorSpaceProposal.md
```diff
@@ -63,56 +104,72 @@ partial interface CanvasRenderingContext2D {
...
-* Images with no color profile, when drawn to the canvas, are assumed to already be in sRGB color space.
-* Unless otherwise explicitly specified by the user, toDataURL/toBlob will produce resources in sRGB color space. If the encoding format supports colorspace tagging or embedded color profiles, the resource will be tagged as being in sRGB color space.
-* Values written to WebGL drawing buffers (e.g, values written to gl_FragColor or the clear color) are defined to be in the output canvas color space.
+* Values written to WebGL backbuffers (e.g, values written to gl_FragColor or the clear color) are in the canvas's color space and linearity.
+* Images with no color profile, when drawn to the canvas, are assumed to be in the sRGB color space.
+* Unless otherwise explicitly specified by the user, toDataURL/toBlob will produce resources in sRGB color space, with unorm8 encoding (matching existing behavior), and "luminance" linearity. If the destination image format supports colorspace tagging or embedded color profiles, the resource will be tagged as being in sRGB color space.
```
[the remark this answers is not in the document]
kenrussell: I had the same question. Should this combination be disallowed, and if so, how? https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext is specified to return the same context if one was fetched previously, no matter whether the context creation attributes are legal, so detecting whether a combination is legal should probably not be delegated to getContext(), but rather to some other API.

CanvasColorSpaceProposal.md
kenrussell: I agree. It's not clear to me how it would interact with a choice of unorm8 vs. unorm8-srgb encoding, for example. How about making the other changes first and deferring addition of the linearity enum?

CanvasColorSpaceProposal.md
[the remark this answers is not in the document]
jdashg: Color space and encoding are technically orthogonal, so that would work fine. That'd be using rec2020 chromaticities (primaries) and (output) transfer function (gamma), but encoding 0.5 as 0xbc. 

Color space includes both chromaticities (primaries and white point) and transfer function, not just RGB primaries, so color space is more accurate IMO. Rec2020 (and friends) do also define a preferred encoding, but we can ignore it for canvas. (it would be prohibitively slow for canvas)

CanvasColorSpaceProposal.md
jdashg: I think this is supported by this PR. (srgb+float16)
What I don't think we want is the srgb-chromaticity+rec2020-trans-func combo from the previous version. That would be an unfortunate subtle deviation from the CSS colors spec.

CanvasColorSpaceProposal.md
jdashg: Happy to do that, though the linearity enum was the replacement for the behavior of the previous version's rec2020 enum, at least to my understanding of it.
I do think that it's non-essential, but I included it for completeness.

CanvasColorSpaceProposal.md
```diff
@@ -26,34 +27,74 @@ Some implementations convert images drawn to canvases to the sRGB color space. T
...
+};
+
+partial dictionary WebGLContextAttributes {
+  CanvasColorSpaceEnum colorSpace = "srgb";
+  CanvasColorEncodingEnum colorEncoding = "unorm8";
+  CanvasColorLinearityEnum colorLinearity = "luminance";
+};
+
+dictionary ImageDataColorSettings {
+  CanvasColorSpaceEnum colorSpace = "srgb";
```
MarkCallow, line 95: You should use CanvasColorPrimariesEnum or something similar. Color space is too widely used to mean the whole combination: primaries, transfer function, white point and black point, e.g, in ICC specs.

And since the sRGB primaries are the same as BT.709 you should have an alternate BT.709 enum value that can be used.

CanvasColorSpaceProposal.md
```diff
@@ -63,56 +104,72 @@ partial interface CanvasRenderingContext2D {
...
 
-#### Common properties of all color spaces
+#### The colorSpace canvas creation parameter
+
+* Color spaces match their respective counterparts as defined in the [CSS colorspaces](https://www.w3.org/TR/css-color-4/#predefined).
 * All input colors (e.g, fillStyle or strokeStyle, and gradient stops) follow the same interpretation as CSS color literals, regardless of canvas color space.
-* Images with no color profile, when drawn to the canvas, are assumed to already be in sRGB color space.
-* Unless otherwise explicitly specified by the user, toDataURL/toBlob will produce resources in sRGB color space. If the encoding format supports colorspace tagging or embedded color profiles, the resource will be tagged as being in sRGB color space.
-* Values written to WebGL drawing buffers (e.g, values written to gl_FragColor or the clear color) are defined to be in the output canvas color space.
+* Values written to WebGL backbuffers (e.g, values written to gl_FragColor or the clear color) are in the canvas's color space and linearity.
```
MarkCallow, line 116: Does this mean you are planning to patch people's shaders to do color transforms - that is one set of primaries to another? The hardware only helps with the "linearity".

CanvasColorSpaceProposal.md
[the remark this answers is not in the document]
MarkCallow: > Color space includes both chromaticities (primaries and white point) and transfer function

Then what on earth is the "linearity" attribute all about?

CanvasColorSpaceProposal.md
MarkCallow: I too do not understand the interaction.

As mentioned earlier in my review (and Alexey mentioned too) I think separating primaries and transfer function will be helpful. For example some proposals for handling wider encodings use sRGB/BT.709 primaries with a linear transfer function. We separated these in the Khronos Data Format Specification> I think it has help simplify things because of the extra flexibility it offers.

CanvasColorSpaceProposal.md
```diff
@@ -63,56 +104,72 @@ partial interface CanvasRenderingContext2D {
...
 
 #### Selecting the best color space match for the user agent's display device
 <pre>
-var colorSpace = window.matchMedia("(color-gamut: rec2020)").matches ? "rec2020" : 
+var colorSpace = window.matchMedia("(color-gamut: rec2020)").matches ? "rec2020" :
     (window.matchMedia("(color-gamut: p3)").matches ? "p3" : "srgb");
 </pre>
 
-#### Selecting the best pixelFormat for the user agent's display device
+#### Selecting the best encoding for the user agent's display device
```
MarkCallow, line 165: Even after reading this far I am still having trouble with "encoding" is this context. Too many years of thinking about sRGB encoding. Now that is truly an encoding of the data. Graphics APIs don't refer to encoding. OpenGL Vulkan, DirectX and Metal all have "formats" not encodings. I think you should revert to pixelFormat here.

CanvasColorSpaceProposal.md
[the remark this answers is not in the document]
lexaknyazev: > That'd be using rec2020 chromaticities (primaries) and (output) transfer function (gamma), but encoding 0.5 as 0xbc.

Sorry, I'm not folowing. Rec.2020 defines its own non-linear transfer function ("gamma" curve) that is identical to Rec.709 but slightly different from sRGB. Are you saying that such configuration would imply applying non-linear transfer twice? What's the use-case?

For documentation purposes, we could split fields even further:
- color primaries, i.e. RGB-CIE relations:
  - sRGB/Rec.709
  - DCI P3
  - Rec.2020
- transfer function (precision redistribution), three distinct options:
  - linear
  - non-linear sRGB
  - non-linear Rec.709 / Rec.2020
- data type
  - unorm8
  - unorm16
  - float16

Of course, not all combinations make practical sense so only few of them should be allowed in the spec. In my opinion, these options should probably be enough:

**Legacy**
- sRGB primaries / linear transfer / unorm8

**Correct 8-bit backbuffer**
- sRGB primaries / sRGB transfer / unorm8 

**Display P3**
- P3 primaries / sRGB transfer / unorm16

**Rec.2020**
- Rec.2020 primaries / Rec.2020 transfer / ??? (10 or 12 bit storage)

**Unbiased HDR**
- Rec.2020 primaries / linear transfer / float16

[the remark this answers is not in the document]
kenrussell: This is a lot of options to provide initially. Can you think of a way to whittle down this list to something more feasible to implement initially, while leaving open the possibility to extend the capability set later? From one browser implementer's standpoint, I have no idea what supporting the P3 and Rec.2020 primaries implies from the browser's compositor and from the hardware's overlay plane support.

Perhaps focus on just enabling the "unorm8-srgb" and "float16" encodings with whatever other knobs are absolutely necessary.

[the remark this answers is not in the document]
MarkCallow: > Legacy
>
>    sRGB primaries / linear transfer / unorm8

I know nothing about canvas 2D implementations but the WebGL canvas is sRGB both in primaries and transfer function (canvas bits to the eyes). The only thing that thinks differently is WebGL which labels it linear thus turning off the GPU's sRGB helpers.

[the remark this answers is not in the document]
lexaknyazev: >  The only thing that thinks differently is WebGL ...

That's what I called "Legacy". From an application POV, the backbuffer doesn't have a transfer function.

>  Can you think of a way to whittle down this list to something more feasible to implement initially,

That's exactly the point of providing fixed presets in well-defined terms instead of independent knobs to a user. This bullets of this list could be implemented step-by-step, as the ecosystem advances.

- "Legacy" - available today
- "Correct 8-bit backbuffer" - highly-anticipated sRGB-correct backbuffer
- "Display P3" - configuration natively supported by the latest high-end mobile devices
- "Rec.2020" - no easy path today, just an example for future
- "Unbiased HDR" - most-flexible, implementation-dependent

CanvasColorSpaceProposal.md
jdashg: I believe we could call this "chromaticity", which iirc is `(primaries, white-point)`.

jdashg: Oh, nevermind, this enum is supposed to include transfer function, because that's required for conversions. I removed the old proposal's support of arbitrary `(chromaticity, transfer-func)` pairs.

CanvasColorSpaceProposal.md
```diff
@@ -3,11 +3,11 @@
 ## Use Case Description
 * Contents displayed through a canvas element should be color managed in order to minimize differences in appearance across browsers and display devices. Improving color fidelity matters a lot for artistic uses (e.g. photo and paint apps) and for e-commerce (product presentation).
 * Canvases should be able to take advantage of the full color gamut and dynamic range of the display device.
-* Some applications prefer that compositing, filtering and interpolation calculations be performed in a physically based linear color space.
```
ccameron-chromium, line 6: Linear-space blending was a goal of the original proposal, but I agree with your removal of it.

CanvasColorSpaceProposal.md
```diff
@@ -6,8 +6,9 @@
 * Some applications prefer that compositing, filtering and interpolation calculations be performed in a physically based linear color space.
 
 ### Current Limitations
-* The color space of canvases is undefined in the current specification.
+* The color space of canvases is undefined in the current specification, though de facto sRGB.
 * The bit-depth of canvases is currently fixed to 8 bits per component, which is below the capabilities of some monitors. Monitors with higher contrast ratios require more bits per component to avoid banding.
+* Color encoding and blending is de facto luminance-linear unorm8, which allocates too little precision on low values (and too much precision on high values), leading to poor handling of dark scenes and content.
```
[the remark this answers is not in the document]
ccameron-chromium: I'd just leave this part out.

CanvasColorSpaceProposal.md
ccameron-chromium: I would include the full list from the color-gamut media query
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut
So that would be srgb, p3, and rec2020.

I think we should also include extended-srgb and linear-extended-srgb as options.

CanvasColorSpaceProposal.md
```diff
@@ -26,34 +27,74 @@ Some implementations convert images drawn to canvases to the sRGB color space. T
...
   "srgb", // default
-  "linear-srgb",
+  "rec-2020",
 };
 
-enum CanvasPixelFormat {
-  "8-8-8-8", // default
+enum CanvasColorEncodingEnum {
+  "unorm8",      // default, 0.5 encoded as 0x80
+  "unorm8-srgb", // 0.5 encoded as 0xbc
```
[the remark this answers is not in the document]
ccameron-chromium: I'd advocate for removing the luminance property. We should probably be careful about the "0.5 encoded as 0x80 vs 0.5 encoded as 0xbc" comment. The actual-bits-stored encoding is the same, it's just that the gl_FragColor and the sampler values are transformed. I'm fine with this as a starting point for iteration, though.

CanvasColorSpaceProposal.md
```diff
@@ -63,56 +89,58 @@ partial interface CanvasRenderingContext2D {
...
 
-#### Common properties of all color spaces
+#### The colorSpace canvas creation parameter
+
+* Color spaces match their respective counterparts as defined in the [CSS colorspaces](https://www.w3.org/TR/css-color-4/#predefined).
 * All input colors (e.g, fillStyle or strokeStyle, and gradient stops) follow the same interpretation as CSS color literals, regardless of canvas color space.
-* Images with no color profile, when drawn to the canvas, are assumed to already be in sRGB color space.
-* Unless otherwise explicitly specified by the user, toDataURL/toBlob will produce resources in sRGB color space. If the encoding format supports colorspace tagging or embedded color profiles, the resource will be tagged as being in sRGB color space.
-* Values written to WebGL drawing buffers (e.g, values written to gl_FragColor or the clear color) are defined to be in the output canvas color space.
+* Values written to WebGL backbuffers (e.g, values written to gl_FragColor or the clear color) are in the canvas's color space.
```
ccameron-chromium, line 100: We'll probably want to call out the subtleties of unorm8-srgb here.

CanvasColorSpaceProposal.md
```diff
@@ -63,56 +89,58 @@ partial interface CanvasRenderingContext2D {
...
+* This color space has different primaries and a different transfer function than "srgb".
+* Support is optional, and should not be present on srgb-only UAs.
+
+
+#### The colorEncoding context creation attribute
+The colorEncoding attributes specifies the encoding to be used for storing pixel channel color values.
+* Support for "unorm8" is mandatory. All other encodings are optional.
+* When an unsupported encoding is requested, the encoding shall fall back to "unorm8".
+* The alpha channel is always interpreted as if clamped to [0,1].
+* Float RGB channel values outside of [0,1] range can be used to represent colors outside of the chosen color gamut. This allows float pixel formats to represent all possible colors and brightness levels. How values outside of [0,1] are displayed depends on the capabilites of the device and output display. Some implementations may simply clamp these values to [0,1]. If the device and display are capable, a (luminance-linear) pixel value of (2,2,2) should be twice as bright as (1,1,1).
```
ccameron-chromium, line 125: I think we should clamp to [0, 1] except for color spaces (e.g, "extended-srgb") which explicitly support values outside the range of [0, 1].

The reason for this is that the extended values can be used for HDR, and enabling HDR often requires falling down a different (higher power) path.

CanvasColorSpaceProposal.md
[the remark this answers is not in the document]
kenrussell: @jdashg 's current draft has removed the luminance property, so this should be resolved now.

CanvasColorSpaceProposal.md
kenrussell: In order to make forward progress on these APIs which are still in the draft stage, could we agree to make the change to encoding / colorEncoding so there's symmetry with colorSpace?

kenrussell:
Thanks @jdashg for taking the initiative to move this proposal forward. Merging now.

kenrussell:
Edited the commit's description during the merge to remove the reference to linearity, and also edited above to match.

PR #30, opened 2020-08-20
Separate IDL sections for Canvas/WebGL/ImageData

ccameron-chromium:
Add separate subsections for the IDL changes for CanvasColorSpace,
CanvasRenderingContext2D, WebGL 2, and ImageData.

Add Display P3 color space enums

Create separate types for the pixel format of CanvasRenderingContext2D, 
WebGL, and ImageData. The supported formats for ImageData are perforce
different (there exists no Float16Array type). Separating WebGL from
CanvasRenderingContext2D is needed because unorm8-srgb are not
substantially meaningful for CanvasRenderingContext2D (they would
have slightly different blending behavior), but are substantially
meaningful for WebGL (changes gl_FragColor interpretation).

Add changes to getImageData and createImageData CanvasRenderingContext2D
APIs.

Add subsections for behavior changes to HTMLCanvasElement.

Add TODOs regarding extended color spaces and HDR support.

Add TODOs regarding UNPACK_COLORSPACE_CONVERSION_WEBGL.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -48,175 +48,150 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
 * Add a canvas context creation attribute to specify an encoding format for storing pixel values.
 * Color space and encoding format parameters are also extended to image storage interfaces that interact with canvases, such as CanvasPattern, ImageData and ImageBitmap.
 
-### Processing Model
+### CanvasColorSpace
 
-IDL:
+IDL Additions:
 <pre>
 // Feature enums:
-
 enum CanvasColorSpaceEnum {
   "srgb", // default
+  "display-p3",
   "rec-2020",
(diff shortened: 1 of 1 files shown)


kenrussell:
I would have liked to see a little more discussion of this pull request.

Specifically, there are several dictionaries now that use different names and enum types for the same concepts ("colorEncoding" vs. "pixelFormat" vs. "storageFormat"). This will lead to cognitive overhead for developers having to remember which one to use where.

`Canvas.getContext()` is a central API where concepts like `alpha`, and their defaults, are unified among the various rendering APIs - at least, 2D canvas and WebGL. I firmly believe that we should continue to aim for a minimal set of unified behavior among these, rather than having slightly different behavior between various APIs like 2D canvas, WebGL and ImageData, and having to think about conversions and compatibility between all of them.

CC'ing @jdashg as well.

PR #32, opened 2020-08-31
Separate IDL sections for Canvas/WebGL/ImageData

ccameron-chromium:
Add separate subsections for the IDL changes for CanvasColorSpace
and CanvasColorEncoding, CanvasRenderingContext2D, WebGL 2, and
ImageData.

Add a pixelFormat context creation attribute for WebGL, in which pixel
formats such as GL_RGBA8 and GL_SRGB8_ALPHA8 may be specified.
This complements the CanvasColorEncoding attribute by allowing
WebGL to indicate the particular GL interpretation of the bits in the
backbuffer.

Add changes to getImageData and createImageData CanvasRenderingContext2D
APIs.

Add subsections for behavior changes to HTMLCanvasElement.

Add TODOs regarding extended color spaces and HDR support.

Add TODOs regarding UNPACK_COLORSPACE_CONVERSION_WEBGL.
Add subsections for behavior changes to HTMLCanvasElement.

Add TODOs regarding extended color spaces and HDR support.

Add TODOs regarding UNPACK_COLORSPACE_CONVERSION_WEBGL.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
 
 * Clearly define the color space of canvases.  Most current browser implementations are either color-managed or are currently actively working on becoming color-managed.  Therefore, it will be possible to specify that the default color space of canvases as 'srgb' instead of leaving it undefined in the spec (as is currently the case due to lack of interoperability).
 * Add a canvas context creation attribute to specify a color space.
-* Add a canvas context creation attribute to specify an encoding format for storing pixel values.
-* Color space and encoding format parameters are also extended to image storage interfaces that interact with canvases, such as CanvasPattern, ImageData and ImageBitmap.
+* Add a canvas context creation attribute to specify a type for the individual color values of the pixels.
+* Color space and color type parameters are also extended to image storage interfaces that interact with canvases, such as CanvasPattern, ImageData and ImageBitmap.
 
-### Processing Model
+### CanvasColorSpace
 
-IDL:
+IDL Additions:
 <pre>
 // Feature enums:
-
 enum CanvasColorSpaceEnum {
   "srgb", // default
+  "display-p3",
   "rec-2020",
 };
 
-enum CanvasColorEncodingEnum {
-  "unorm8",      // default, 0.5 encoded as 0x80
-  "unorm8-srgb", // 0.5 encoded as 0xbc
-  "float16",     // IEEE 754
+enum CanvasColorTypeEnum {
+  "uint8", // default
+  "float16",
 };
 
 // Feature detection:
-
 interface CanvasColorSpace {
   const CanvasColorSpaceEnum srgb = "srgb";
+  const CanvasColorSpaceEnum displayP3 = "display-p3";
   const CanvasColorSpaceEnum rec2020 = "rec-2020";
 };
 
-interface CanvasColorEncoding {
-  const CanvasColorEncodingEnum unorm8 = "unorm8";
-  const CanvasColorEncodingEnum unorm8Srgb = "unorm8-srgb";
-  const CanvasColorEncodingEnum float16 = "float16";
+interface CanvasColorType {
+  const CanvasColorTypeEnum uint8 = "uint8";
+  const CanvasColorTypeEnum float16 = "float16";
 };
+</pre>
 
-// Feature activation:
+Color spaces match their respective counterparts as defined in the [CSS colorspaces](https://www.w3.org/TR/css-color-4/#predefined). Support for color spaces should not be limited based on the capabilities of the user's display. There exist [Media Query APIs](https://www.w3.org/TR/mediaqueries-4/) for that purpose.
 
+When displaying a canvas element, the browser must ensure that the colors displayed on the user's screen match the colors specified in the canvas element as closely as is possible on the platform. The color space conversions responsible for this happen in compositing, and should have no script-visible side-effects.
+
+Some colorType values will allow the user to specify floating-point colors outside of the usual [0, 1] interval. These values will be clamped to the range [0, 1] when being displayed.
+
+TODO: Add 'extended-srgb' and 'extended-srgb-linear' color spaces which will not be subject to this constraint.
+
+TODO: Add a mechanism for indicating that extended color spaces are to be used for HDR rendering (this should probably not be the default, because HDR rendering has substantial power costs).
+
+### HTMLCanvasElement API Changes
+
+The behavior of the HTMLCanvasElement functions toDataURL and toBlob methods are to produce encodings that match the canvas element's color space and pixel depth as closely as possible, subject to the limitations of the encoding format. It may be appropriate to add additional parameters to the relevant APIs to allow for more user control in this area.
+
+### CanvasRenderingContext2D
+
+IDL Additions:
+<pre>
 partial dictionary CanvasRenderingContext2DSettings {
-  CanvasColorSpaceEnum colorSpace = "srgb";
-  CanvasColorEncodingEnum colorEncoding = "unorm8";
-};
-
-partial dictionary WebGLContextAttributes {
-  CanvasColorSpaceEnum colorSpace = "srgb";
-  CanvasColorEncodingEnum colorEncoding = "unorm8";
-};
-
-dictionary ImageDataColorSettings {
-  CanvasColorSpaceEnum colorSpace = "srgb";
-  CanvasColorEncodingEnum colorEncoding = "unorm8";
+  CanvasColorSpace colorSpace = "srgb";
+  CanvasColorType colorType = "uint8";
 };
 
 partial interface CanvasRenderingContext2D {
   CanvasRenderingContext2DSettings getContextAttributes();
+  ImageData getImageData(long sx, long sy, long sw, long sh, optional ImageDataSettings imageDataSettings);
+  ImageData createImageData(long sw, long sh, optional ImageDataSettings imageDataSettings);
 };
 </pre>
 
-Example:
+All input colors (e.g, fillStyle or strokeStyle, and gradient stops) follow the same interpretation as CSS color literals, regardless of canvas color space. Images that do not specify a color space, when drawn to the canvas, are treated as though they specified the sRGB color space.
+
+Note that because there exists no direct access to the canvas backbuffer, it is not required that the colorType be truly respected. From the user's point of view, there will be no way to know if the true backing has higher precision than what was requested (e.g, the canvas is actually float32, when float16 was specified).
+
+### WebGL 2 API Changes
+
+IDL Additions:
 <pre>
-canvas.getContext('2d', { colorSpace: "rec2020",
-                          colorEncoding: "float16"} );
+partial dictionary WebGLContextAttributes {
+  CanvasColorSpace colorSpace = "srgb";
+  CanvasColorType colorType = "uint8";
+  GLenum pixelFormat = GL_NONE;
+};
 </pre>
 
-#### The colorSpace canvas creation parameter
+Values stored in the WebGL backbuffers  are in the canvas's color space, as specified by the colorSpace context attribute.
 
-* Color spaces match their respective counterparts as defined in the [CSS colorspaces](https://www.w3.org/TR/css-color-4/#predefined).
-* All input colors (e.g, fillStyle or strokeStyle, and gradient stops) follow the same interpretation as CSS color literals, regardless of canvas color space.
-* Values written to WebGL backbuffers (e.g, values written to gl_FragColor or the clear color) are in the canvas's color space.
-* Images with no color profile, when drawn to the canvas, are assumed to be in the sRGB color space.
-* Unless otherwise explicitly specified by the user, toDataURL/toBlob will produce resources in sRGB color space, with unorm8 encoding (matching existing behavior). If the destination image format supports colorspace tagging or embedded color profiles, the resource will be tagged as being in sRGB color space.
+The pixelFormat specifies the WebGL format of the backbuffer. If a value of GL_NONE is specified, then a default pixelFormat is determined based on the colorType and alpha attributes. The following table lists the supported combinations of colorFormat, alpha, and pixelFormat. The default pixelFormat for a given colorType and alpha pair is indicated by (default). All combinations not listed in this table are invalid.
 
-##### The "srgb" color space
-* This color space matches the existing canvas behavior.
-* Guarantees that color values used as fillStyle or strokeStyle exactly match the appearance of the same color value when it is used in CSS.
-* On implementations that do not color-manage CSS colors, the canvas "srgb" color space must not be color-managed either, in order to preserve color-matching between CSS and canvas-rendered content. This situation shall be referred to as the "legacy behavior".
-* All content drawn into such a 2d canvas RC must be color corrected to sRGB.
-    * Exception: User agents that implement the legacy behavior must apply color correction steps that match the color correction that is applied to image resources that are displayed via CSS.
-* Displayed canvases must be color corrected for the display if a display color profile is available. This color correction happens downstream at the compositing stage, and has no script-visible side-effects.
-* toDataURL/toBlob produce resources tagged as being in the sRGB color space, if the encoding format supports colorspace tagging or embedded color profiles.
-    * Exception: User agents that implement the legacy behavior must not encode any color space metadata.
+| colorType | alpha | Implied pixelFormat |
+| - | - | - |
+| "uint8" | false | RGB8 (default) |
+| "uint8" | false | SRGB8 |
+| "uint8" | true | RGBA8 (default) |
+| "uint8" | true | SRGB8_ALPHA8 |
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
In this patch I've added an explicit WebGL pixel format type (e.g, gl.RGB8, gl.SRGB8_ALPHA8) to the WebGL context creation parameters. I've kept the existing CanvasColorEncoding (renamed to CanvasColorType) along with the WebGL context creation parameter, but added a TODO suggesting that we remove one of them

I think that ultimately WebGL's pixel format specification will need to be distinct from 2D canvas. 2D canvas can't meaningfully support all formats (e.g, gl.SRGB8_ALPHA) in the same way that WebGL can. Also, 2D canvas has a lot of flexibility about what it does behind the scenes (e.g, one could request 10 bits per channel, and 2D canvas could lie and just use 16 bits per channel without much consequence, while for WebGL those details can leak through).

I renamed CanvasColorEncoding to CanvasColorType, but I'm still not happy with the name -- suggestions very welcome. It may be that we should make it so that they attribute keys for it and ImageDataStorageFormat be the same (e.g, CanvasStorageFormat).

It also may be that the WebGL changes should be moved into a two separate WebGL extensions. One to allow specifying the pixel format of the backbuffer, and another to add a pile of UNPACK_COLORSPACE_CONVERSION_WEBGL enums. In that case, the only WebGL change here would be the color space in which the compositor interprets WebGL pixel values (sRGB v P3, etc).

kenrussell:
Thanks @ccameron-chromium for re-uploading this PR!

Repo owners (@jdashg , @fserb, others), let's discuss these changes before merging them.

ccameron-chromium:
Was there any further feedback on this PR?

fserb:
LGTM

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
+IDL Additions:
+<pre>
+partial dictionary CanvasRenderingContext2DSettings {
+  CanvasColorSpace colorSpace = "srgb";
+  CanvasColorType colorType = "uint8";
 };
 
 partial interface CanvasRenderingContext2D {
   CanvasRenderingContext2DSettings getContextAttributes();
+  ImageData getImageData(long sx, long sy, long sw, long sh, optional ImageDataSettings imageDataSettings);
```
RafaelCintron, line 105: I do not see `ImageDataSettings` mentioned anywhere in this document besides this IDL definition.  

Please add a definition in this document along with a section on why it is needed and the behavior of `getImageData/createImageData` when passed the settings object.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
+partial dictionary WebGLContextAttributes {
+  CanvasColorSpace colorSpace = "srgb";
+  CanvasColorType colorType = "uint8";
+  GLenum pixelFormat = GL_NONE;
+};
+</pre>
 
-#### Feature detection
-2D rendering contexts are to expose a new getContextAttributes() method, that works much like the method of the same name on WebGLRenderingContext. The method returns the "actual context attributes" which represents the settings that were successfully applied at context creation time. The settings attribute reflects the result of running the algorithm for coercing the settings argument for 2D contexts, as well as the result of any fallbacks that may have happened as a result of options not being supported by the UA.
+Values stored in the WebGL backbuffers  are in the canvas's color space, as specified by the colorSpace context attribute.
```
RafaelCintron, line 125: Nit: Extra space between the words "backbuffers" and "are".

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
 Authors of games and imaging apps are expected to be enthusiastic adopters.
 
-## Unresolved Issues
+## Issues
 
-* Should we support custom color spaces based on ICC profiles? Would offer ultimate flexibility. Would be hard to make implementations as efficient as built-in color spaces, in particular for implement linearPixelMath for profiles that have arbitrary transfer curves.
+Issues:
 
-* Should float16 allow alpha values outside [0,1] range at any stage of the pipeline? What would they mean?
+* How should feature detection work? The combinations of CanvasColorSpace and CanvasColorType supported by 2D Canvas and WebGL are, for now, obviously "all of them", but this is not likely to remain the case.
```
RafaelCintron, line 187: The canvas context attributes can already be queried by the user after the canvas context has been created.  If the query returns a dictionary with no `colorSpace` or `storageFormat` members, the developer can conclude they were ignored.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
 ### Adoption
+
 Lack of color management and color interoperability is a longstanding complaint about the canvas API.
 Authors of games and imaging apps are expected to be enthusiastic adopters.
 
-## Unresolved Issues
+## Issues
 
-* Should we support custom color spaces based on ICC profiles? Would offer ultimate flexibility. Would be hard to make implementations as efficient as built-in color spaces, in particular for implement linearPixelMath for profiles that have arbitrary transfer curves.
+Issues:
```
RafaelCintron, line 185: How will a web developer determine whether advanced color management is necessary?  

On an SDR monitor, using the proposed options incurs unnecessary texture space (for float16 textures) and processing time to tonemap from the source to destination color gamuts/spaces.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
+The pixelFormat specifies the WebGL format of the backbuffer. If a value of GL_NONE is specified, then a default pixelFormat is determined based on the colorType and alpha attributes. The following table lists the supported combinations of colorFormat, alpha, and pixelFormat. The default pixelFormat for a given colorType and alpha pair is indicated by (default). All combinations not listed in this table are invalid.
 
-Note: An alternative approach that was considered was to augment the probablySupportsContext() API by making it check the second argument.  That approach is difficult to consolidate with how dictionary arguments are meant to work, where unsupported entries are just ignored.
+| colorType | alpha | Implied pixelFormat |
+| - | - | - |
+| "uint8" | false | RGB8 (default) |
+| "uint8" | false | SRGB8 |
+| "uint8" | true | RGBA8 (default) |
+| "uint8" | true | SRGB8_ALPHA8 |
+| "float16" | true | RGBA16F |
```
RafaelCintron, line 135: If the web developer creates a canvas with `pixelFormat` of `GL_NONE` (or doesn't include it in the first place) and subsequently queries the context attributes after creation, will the `pixelFormat` member be `GL_NONE` or the value chosen from the table?  

Might be best to keep things simple and force developers to specify a `pixelFormat` from the table.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
-
-#### Selecting the best color space match for the user agent's display device
-<pre>
-var colorSpace = window.matchMedia("(color-gamut: rec2020)").matches ? "rec2020" :
-    (window.matchMedia("(color-gamut: p3)").matches ? "p3" : "srgb");
-</pre>
-
-#### Selecting the best encoding for the user agent's display device
-Selection should be based on the best color space match (see above). For srgb, at least 8 bits per component is recommended; for p3, 10 bits; and for rec2020, 12 bits.  The float16 format is suitable for any colorspace.  There may soon be a proposal to add a way of detecting HDR displays, possibly something like "window.screen.isHDR()" (TBD), which would be a good hint to use the float16 format.
+Note that because there exists no direct access to the canvas backbuffer, it is not required that the colorType be truly respected. From the user's point of view, there will be no way to know if the true backing has higher precision than what was requested (e.g, the canvas is actually float32, when float16 was specified).
```
RafaelCintron, line 112: When would the canvas be float32?

RafaelCintron (changes requested):
Would be good to separate this PR into a cleanup one and a different one which adds functionality.  The title is misleading as it's doing more than just making separate IDL sections.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
-For future consideration: support could be added for color space defined using the [CSS @color-profile rule](https://www.w3.org/TR/css-color-4/#at-profile).
+### WebGL 2 API Changes
 
-#### Compositing the canvas element
-Canvas contents are composited in accordance with the canvas element's style (e.g. CSS compositing and blending rules). The necessary compositing operations must be performed in an intermediate colorspace, the compositing space, that is implementation specific. The compositing space must have sufficient precision and a sufficiently wide gamut to guarantee no undue loss of precision or gamut clipping in bringing the canvas's contents to the display. Implementations should not expose color spaces that are unreasonble for the display.
+IDL Additions:
+<pre>
+partial dictionary WebGLContextAttributes {
+  CanvasColorSpace colorSpace = "srgb";
+  CanvasColorType colorType = "uint8";
```
jdashg, line 120: `unorm8` is more precise, and I would prefer it. Propose this separately.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
+  GLenum pixelFormat = GL_NONE;
```
jdashg, line 121: Propose this separately. What's the usecase for decoupling this from colorSpace/Type?

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
-TODO(ccameron): Review this
+### ImageData API Changes
 
-IDL
+IDL Additions:
 <pre>
+typedef (Uint8ClampedArray or Uint16Array or Float32Array) ImageDataArray;
 
-enum ImageDataStorageType {
+enum ImageDataStorageFormat {
```
jdashg, line 149: `Type` is more consistent than `Format`, given the overloads of "format" in the context of WebGL.

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
-  readonly attribute unsigned long height;
-  readonly attribute ImageDataArray data;
+dictionary ImageDataAttributes {
+  CanvasColorSpace colorSpace = "srgb";
+  ImageDataStorageFormat storageFormat = "uint8";
+};
 
-  ImageDataColorSettings getColorSettings();
+partial interface ImageData {
+  ImageDataAttributes getAttributes();
```
jdashg, line 161: Why not use a readonly attribute?

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
-* Should we support custom color spaces based on ICC profiles? Would offer ultimate flexibility. Would be hard to make implementations as efficient as built-in color spaces, in particular for implement linearPixelMath for profiles that have arbitrary transfer curves.
+Issues:
 
-* Should float16 allow alpha values outside [0,1] range at any stage of the pipeline? What would they mean?
+* How should feature detection work? The combinations of CanvasColorSpace and CanvasColorType supported by 2D Canvas and WebGL are, for now, obviously "all of them", but this is not likely to remain the case.
 
-* Should there be API-level support for mixing chromaticities and transfer functions, including use of no-op transfer functions?
+* Should we unify ImageDataStorageFormat with CanvasColorType? This might create more hassle than it's worth, because there are types for CanvasColorType (e.g, float16) that cannot exist for ImageDataStorageFormat at this time.
+
+* For ImageData, is it appropriate to be able to query the ImageDataStorageFormat, when that can be inferred from the type of the data member? This would make sense if we supported both uint8 and unorm8.
```
jdashg, line 191: What is `uint8` that isn't actually `unorm8`?

CanvasColorSpaceProposal.md
```diff
@@ -45,178 +45,172 @@ You can see this by making stripes of 1.0 and 0.0, and matching that pattern's b
...
+### CanvasColorSpace
 
-IDL:
+IDL Additions:
 <pre>
 // Feature enums:
-
 enum CanvasColorSpaceEnum {
   "srgb", // default
+  "display-p3",
```
jdashg, line 58: Add this separately

jdashg (changes requested):
The title really needs to be fixed.

CanvasColorSpaceProposal.md
ccameron-chromium: 2D canvas can't meaningfully support all formats (e.g, gl.SRGB8_ALPHA) in the same way that WebGL can. Also, 2D canvas has a lot of flexibility about what it does behind the scenes, because 2D canvas doesn't expose direct access to the framebuffer (e.g, one could request 10 bits per channel, and 2D canvas could lie and just use 16 bits per channel without much consequence, while for WebGL those details can leak through).

CanvasColorSpaceProposal.md
ccameron-chromium: Sounds good, I'll use the Type name in a separate patch that just addresses ImageData issues.

CanvasColorSpaceProposal.md
ccameron-chromium: (because I didn't know it was an option :)

CanvasColorSpaceProposal.md
ccameron-chromium: Integer-based texture formats (GL_RGBA8I, MTLPixelFormatRGBA8Uint, etc).

From the perspective of someone using 2D canvas, "unorm" is a foreign concept. In WebGL, it makes sense.

But I'll do this separately.

CanvasColorSpaceProposal.md
ccameron-chromium: Copy.

CanvasColorSpaceProposal.md
ccameron-chromium: I will create a separate patch on ImageDataSettings (it was there in the past, maybe it got lost in a shuffle).

CanvasColorSpaceProposal.md
ccameron-chromium: If a browser were to add, say, "adobe-rgb" as a color space, how do does one detect ahead of time that that is supported? It would be nice to be able to check before creating the context.

(But I'm new in this area, so maybe that's the convention)

CanvasColorSpaceProposal.md
ccameron-chromium: The color gamut media queries can be used to determine if there is a benefit to using p3 or rec2020.
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut

(A similar media query should be added for HDR, but hasn't, to my knowledge).

CanvasColorSpaceProposal.md
ccameron-chromium: I'll remove the GL_NONE part. The reason I added GL_NONE as a default value is because I thought a default value was required, and there is no default value that works universally.

CanvasColorSpaceProposal.md
ccameron-chromium: If the implementation, behind the scenes, prefers float32 (for any implementor reason). That's a bit more of a theoretical example, but as we add formats, more realistic examples may come up. For instance, a user could request 10-bit color, and some implementations (due to driver bugs, etc), may prefer to just use 16-bit color.

CanvasColorSpaceProposal.md
ccameron-chromium: I'd prefer that the unorm8 or uint8 type be removed entirely, in favor of GLenum (e.g, SRGB8_ALPHA8).

ccameron-chromium:
I'm going to turn this into a bunch of separate patches. This proposal needs a serious amount of revision. I had hoped that we could make sweeping changes like this instead of slicing them into smaller proposals, but I can go with the smaller proposals.

CanvasColorSpaceProposal.md
RafaelCintron: @ccameron-chromium , you're right that there's no way to tell ahead of time whether there is support. 
 You have to try things and see whether the request was honored.  

ImageBitmap has [a similar problem with `ImageBitmapOptions`](https://github.com/whatwg/html/pull/4248) which hasn't been resolved.

ccameron-chromium: Thanks for the context!

ccameron-chromium:
I've split out two of the smaller features into new pulls at
https://github.com/WICG/canvas-color-space/pull/33 (very small, just p3)
https://github.com/WICG/canvas-color-space/pull/34 (slightly larger, ImageData and a bit of ImageBitmap)

PR #33, opened 2020-09-10
Add display-p3 as a color space option

ccameron-chromium:
Add "display-p3" as a color space option, addition to "srgb" and "rec-2020".

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -220,7 +220,7 @@ Authors of games and imaging apps are expected to be enthusiastic adopters.
 
 * Should context creation throw on an unrecognized, non-undefined creation attribute?
 
-* The [Media Query APIs](https://www.w3.org/TR/mediaqueries-4/) use the names "p3" and "rec2020" instead of "display-p3" and "rec-2020". This divergence could be confusing.
+* The [Media Query APIs](https://www.w3.org/TR/mediaqueries-4/) use the names "p3" and "rec2020", while the [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#predefined) uses the names "display-p3" and "rec-2020". This divergence could be confusing.
 
 ## Proposal History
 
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
Any comments on this?

CanvasColorSpaceProposal.md
```diff
@@ -218,6 +220,8 @@ Authors of games and imaging apps are expected to be enthusiastic adopters.
 
 * Should context creation throw on an unrecognized, non-undefined creation attribute?
 
+* The [Media Query APIs](https://www.w3.org/TR/mediaqueries-4/) use the names "p3" and "rec2020" instead of "display-p3" and "rec-2020". This divergence could be confusing.
```
kenrussell, line 223: Would this be a good opportunity to align them? If this pull request added "p3" instead of "display-p3" it would reduce divergence.

kenrussell (approved):
Seems straightforward. I assume this will involve more work for the display compositor. Are there subtle side-effects of this, like changing the representation of the canvas's backing store?

CanvasColorSpaceProposal.md
ccameron-chromium: Both "p3" and "display-p3" have left the barn (WebKit ships it for CSS colors).

I'd prefer to align with CSS color spec. If we change that, then I'm open to changing this.

kenrussell: Could you add a reference to the CSS color spec then to make it clearer that this aligns with that spec, and the Media Query APIs are the odd one out?

ccameron-chromium: Updated with a link to the CSS Color Level 4.

kenrussell (approved):
Thanks, looks good to me!

jdashg (approved):
I think rec2020 and p3 are better names, but I'm not zealous about it.

ccameron-chromium:
> I think rec2020 and p3 are better names, but I'm not zealous about it.

Yeah, I could see it going either way -- for now I'm aligning with the CSS color spec, but maybe that spec should change. WebKit does allow specifying CSS colors as "display-p3" already.

PR #34, opened 2020-09-10
Update ImageData and ImageBitmap sections

ccameron-chromium:
Move the ImageDataColorSettings definition to the ImageData section of the specification.

Update ImageDataSettings to use the existing ImageDataStorageType type, and rename the ImageDataStorageTypes of uint8 and uint16 to unorm8 and unorm16.

Change the default ImageDataSettings for createImageData and getImageData to be their current values of uint8 and srgb. This is preferable because it is more well-defined (there is no implicit selection of the "best matching" ImageDataStorageType) and because it preserves existing behavior.

Add text indicating that ImageBitmap will make its best effort to preserve the fidelity of what it is representing.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -90,11 +90,6 @@ partial dictionary WebGLContextAttributes {
   CanvasColorEncodingEnum colorEncoding = "unorm8";
 };
 
-dictionary ImageDataColorSettings {
-  CanvasColorSpaceEnum colorSpace = "srgb";
-  CanvasColorEncodingEnum colorEncoding = "unorm8";
-};
-
 partial interface CanvasRenderingContext2D {
   CanvasRenderingContext2DSettings getContextAttributes();
 };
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
Any feedback on this?

CanvasColorSpaceProposal.md
```diff
@@ -164,40 +159,55 @@ Note: An alternative approach that was considered was to augment the probablySup
...
-IDL
+Add the following types to be used by `ImageData`.
 <pre>
-
-enum ImageDataStorageType {
-  "uint8", // default
-  "uint16",
+enum ImageDataEncoding {
+  "unorm8", // default
+  "unorm16",
```
kenrussell, line 170: What's the motivation behind this change? It introduces a new concept which doesn't align with the Typed Array specification - these will be accessed with Uint8Array and Uint16Array. I would prefer to leave these as "uint8" and "uint16", and define elsewhere that these are treated as unsigned normalized values.

CanvasColorSpaceProposal.md
```diff
@@ -164,40 +159,55 @@ Note: An alternative approach that was considered was to augment the probablySup
...
 </pre>
 
-* When using the constructor that takes an ImageDataArray parameter, the "storageType" setting is ignored.
-* createImageData() and getImageData() produce an ImageData object with the same color space as the source canvas, using an ImageDataArray of a type that is appropriate for the pixelFormat of the source canvas (smallest possible numeric size that guarantees no loss of precision).
-* putImageData() performs a color space conversion to the color space of the destination canvas.
-* Data returned by getImageData() or passed to putImageData() are assumed to be in "srgb" color space, with "unorm8" encoding.
+The changes to this interface are:
+* The constructors now take an optional `ImageDataSettings` dictionary.
+* The ImageDataSettings attribute may be queried using `getImageDataSettings`.
+* The constructor and attribute that used to be a `Uint8ClampedArray` are now a `ImageDataArray` union, which can specify data in multiple formats.
```
kenrussell, line 193: Please add a table specifying the mapping between the ImageDataEncoding enum values and the associated typed array types.

kenrussell (commented):
Sorry for the delay reviewing. A couple of comments.

CanvasColorSpaceProposal.md
ccameron-chromium: I'm agnostic. Theoretically, one could request an actual uint (as in, integer texture) encoding, but that's probably just a footgun. I've changed it back.

CanvasColorSpaceProposal.md
ccameron-chromium: Updated.

CanvasColorSpaceProposal.md
```diff
@@ -164,43 +159,67 @@ Note: An alternative approach that was considered was to augment the probablySup
...
 #### ImageData
 
-TODO(ccameron): Review this
-
-IDL
+Add the following types to be used by `ImageData`.
 <pre>
-
-enum ImageDataStorageType {
+enum ImageDataEncoding {
```
kenrussell, line 168: Noting that there was an accidental reference to the old "ImageDataStorageType" enum below, I wonder whether this change from "storage type" to "encoding" is warranted. There's precedent in the OpenGL API for the term "type" and in the Direct3D API for "format". Since JavaScript uses the term "Typed Arrays" I wonder whether this should just be left as "ImageDataStorageType" for symmetry.

CanvasColorSpaceProposal.md
```diff
@@ -164,43 +159,67 @@ Note: An alternative approach that was considered was to augment the probablySup
...
   "uint8", // default
   "uint16",
   "float32",
 };
-
+dictionary ImageDataSettings {
+  CanvasColorSpaceEnum colorSpace = "srgb";
+  ImageDataStorageType encoding = "uint8";
```
kenrussell, line 175: This still references the "ImageDataStorageType" data type. See above discussion. I lean toward leaving it as "ImageDataStorageType".

CanvasColorSpaceProposal.md
```diff
@@ -164,43 +159,67 @@ Note: An alternative approach that was considered was to augment the probablySup
...
-* When using the constructor that takes an ImageDataArray parameter, the "storageType" setting is ignored.
-* createImageData() and getImageData() produce an ImageData object with the same color space as the source canvas, using an ImageDataArray of a type that is appropriate for the pixelFormat of the source canvas (smallest possible numeric size that guarantees no loss of precision).
-* putImageData() performs a color space conversion to the color space of the destination canvas.
-* Data returned by getImageData() or passed to putImageData() are assumed to be in "srgb" color space, with "unorm8" encoding.
+The changes to this interface are:
+* The constructors now take an optional `ImageDataSettings` dictionary.
+* The ImageDataSettings attribute may be queried using `getImageDataSettings`.
+* The constructor and attribute that used to be a `Uint8ClampedArray` are now a `ImageDataArray` union, which can specify data in multiple formats.
+
+The type of the ``data`` attribute is determined by the ``encoding`` parameter according to the following table.
```
kenrussell, line 195: Depending on resolution of above feedback, consider changing "encoding" to "type" or "storageType".

CanvasColorSpaceProposal.md
ccameron-chromium: Updated all instances to ImageDataStorageType

CanvasColorSpaceProposal.md
ccameron-chromium: Updated all instances to ImageDataStorageType

CanvasColorSpaceProposal.md
ccameron-chromium: Updated to storageType

CanvasColorSpaceProposal.md
```diff
@@ -164,43 +159,67 @@ Note: An alternative approach that was considered was to augment the probablySup
...
 };
 </pre>
 
-* When using the constructor that takes an ImageDataArray parameter, the "storageType" setting is ignored.
-* createImageData() and getImageData() produce an ImageData object with the same color space as the source canvas, using an ImageDataArray of a type that is appropriate for the pixelFormat of the source canvas (smallest possible numeric size that guarantees no loss of precision).
-* putImageData() performs a color space conversion to the color space of the destination canvas.
-* Data returned by getImageData() or passed to putImageData() are assumed to be in "srgb" color space, with "unorm8" encoding.
+The changes to this interface are:
+* The constructors now take an optional `ImageDataSettings` dictionary.
+* The ImageDataSettings attribute may be queried using `getImageDataSettings`.
```
kenrussell, line 192: I guess the only remaining question is whether this should be a readonly attribute or a method. Right now the interface contains only attributes.

kenrussell (approved):
Thanks for the discussion and the updates! Looks good to me.

CanvasColorSpaceProposal.md
ccameron-chromium: I'll have to defer to people who are more expert than me (I don't have enough experience in these APIs to have a sense as to what is natural)

ccameron-chromium: I'll go ahead and merge this, but we should continue to revise it.

PR #35, opened 2020-10-09
Remove older details

ccameron-chromium:
This spec is several years old and has accumulated some cruft. Clear some of it out. Also, remove extra background information and definitions. This information is helpful, but is currently cluttering the spec, and we are trying to pin down the behavior changes.

In particular:
* From "Current Limitations", remove remarks about perceptual-linear unorm8 encoding, because the stored encoding of 8-bit contents is unchanged by this specification (unorm8-srgb affects only blending and fragment math space).
* Remove "Current Usage and Workarounds" section. Most of the behavior documented in this section is obsolete.
* Shorten "Requests for this Feature" section to not specify particular vendor applications.
* Remove the "Background" section and replace with a "Related Specifications" section. Do not define the color spaces in this document, rather, link to the CSS Color Level 4 specification.
* Move the "Selecting the best color space match for the user agent's display device" to an "Examples" section at the end. Remove speculation about a TBD isHDR method.
* Remove discussions of "srgb" and "rec-2020" color spaces.
* Separate subsections for "2D canvas behavior" versus "WebGL behavior", since the effects of the spec are different on the two.
* Move discussion of srgb-encoded pixel values to WebGL behavior section.
* Clarify that undefined or unsupported color spaces should default to srgb with unorm8.
* Remove the advice that "Implementations should not expose color spaces that are unreasonable for the display".

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -25,7 +25,7 @@
 ## Proposed Solution
 
 * Clearly define the default color space of canvases to be ``srgb``, and the default encoding to be 8 bits per pixel.
-* Add canvas context creation attributes to change the color space and encoding of a canvas.
+* Add canvas context creation attributes to specify the color space and encoding of a canvas.
 * Clarify behavior for drawing into canvases, compositing canvases, and exporting canvas contents.
 * Add mechanisms to specify the color space and encoding of ``ImageData`` objects.
 
@@ -111,7 +111,7 @@ The ``colorEncoding`` attribute specifies the encoding to be used for storing pi
 
 * Values stored in WebGL backbuffers are in the canvas's color space.
 * Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
-* For the color encoding of ``"unorm-srgb8"``, the interpretation of values written by ``gl_FragColor`` is as follows:
+* For the color encoding of ``"unorm8-srgb"``, the interpretation of values written by ``gl_FragColor`` is as follows:
   * The color value is in a linear color space.
   * The inverse sRGB transfer function is applied only after blending has occurred.
   * E.g, if the value ``0.5`` is written in ``gl_FragColor``, the result will be stored as ``0xbc``.
@@ -119,7 +119,7 @@ The ``colorEncoding`` attribute specifies the encoding to be used for storing pi
 
 #### Compositing the canvas element
 
-Canvas contents are composited in accordance with the canvas element's style (e.g. CSS compositing and blending rules). The necessary compositing operations must be performed in an intermediate colorspace, the compositing space, that is implementation specific. The compositing space must have sufficient precision and a sufficiently wide gamut to guarantee no undue loss of precision or gamut clipping in bringing the canvas's contents to the display.
+Canvas contents are composited in accordance with the canvas element's style (e.g. CSS compositing and blending rules). The necessary compositing operations must be performed in an intermediate colorspace, the compositing space, that is implementation specific. The compositing space must have sufficient precision and a sufficiently wide gamut to guarantee no undue loss of precision or gamut clipping in bringing the canvas's contents to the display. Implementations should not expose color spaces that are unreasonble for the display.
 
 #### Feature detection
 
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -111,7 +111,7 @@ The ``colorEncoding`` attribute specifies the encoding to be used for storing pi
 
 * Values stored in WebGL backbuffers are in the canvas's color space.
 * Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
-* For the color encoding of ``"unorm-srgb"``, the interpretation of values written by ``gl_FragColor`` is as follows:
+* For the color encoding of ``"unorm-srgb8"``, the interpretation of values written by ``gl_FragColor`` is as follows:
   * The color value is in a linear color space.
   * The inverse sRGB transfer function is applied only after blending has occurred.
   * E.g, if the value ``0.5`` is written in ``gl_FragColor``, the result will be stored as ``0xbc``.
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -111,7 +111,7 @@ The ``colorEncoding`` attribute specifies the encoding to be used for storing pi
 
 * Values stored in WebGL backbuffers are in the canvas's color space.
 * Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
-* The encodings for specific color value differ between ``"unorm8"`` and ``"unorm8-srgb"``.
+* The encodings for specific color values differ between ``"unorm8"`` and ``"unorm8-srgb"``.
 * For ``"unorm8"``, the color encoding function is:
 <pre>
 function encodeUnorm8(val) { return val * 0xff; }
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -3,6 +3,7 @@
 ## Use Case Description
 * Contents displayed through a canvas element should be in a well-defined color space to minimize differences in appearance across browsers and display devices.
 * Canvases should be able to take advantage of the full color gamut and dynamic range of the display device.
+* Contents displayed through a canvas element should be color managed in order to minimize differences in appearance across browsers and display devices. Improving color fidelity matters for artistic (e.g, photo and paint apps) and for e-commerce (e.g, product presentation) use cases.
 
 ### Current Limitations
 * The color space of canvases is undefined in the current specification, although is de facto sRGB.
@@ -13,6 +14,9 @@
 
 * <cite>[https://github.com/whatwg/html/issues/299]</cite> <blockquote><p>Allow 2dcontexts to use deeper color buffers</p></blockquote>
 * <cite>[https://bugs.chromium.org/p/chromium/issues/detail?id=425935]</cite> <blockquote><p>Wrong color profile with 2D canvas</p></blockquote>
+* Engineers from the Google Photos, Maps and Sheets teams have expressed a desire for canvases to be color managed. Particularly for the use case of resizing an image, using a canvas, prior to uploading it to the server, to save bandwidth. The problem is that the images retrieved from a canvas are in an undefined color space and no color space information is encoded by toDataURL or toBlob.
+
+
 
 ### Related Specifications
 * The [CSS Media Queries Level 5](https://www.w3.org/TR/mediaqueries-5/#color-gamut) specification defines the ``color-gamut`` media query to determine the capabilities of the current display device.
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -1,52 +1,29 @@
 # Color managing canvas contents
 
 ## Use Case Description
-* Contents displayed through a canvas element should be color managed in order to minimize differences in appearance across browsers and display devices. Improving color fidelity matters a lot for artistic uses (e.g. photo and paint apps) and for e-commerce (product presentation).
+* Contents displayed through a canvas element should be in a well-defined color space to minimize differences in appearance across browsers and display devices.
 * Canvases should be able to take advantage of the full color gamut and dynamic range of the display device.
 
 ### Current Limitations
-* The color space of canvases is undefined in the current specification, though de facto sRGB.
-* The bit-depth of canvases is currently fixed to 8 bits per component, which is below the capabilities of some monitors. Monitors with higher contrast ratios require more bits per component to avoid banding.
-* Color encoding and blending is de facto perceptually-linear unorm8, which allocates too little precision on low values (and too much precision on high values), leading to poor handling of dark scenes and content.
-
-### Current Usage and Workarounds
-The lack of color space interoperability is hard to work around. For some browser implementations which color-correct images drawn to canvases by applying the display profile, apps that want to use canvases for color corrected image processing are stuck doing convoluted workarounds, such as:
-* reverse-engineer the display profile by drawing test pattern images to the canvas and inspecting the color corrected result via getImageData
-* bypass CanvasRenderingContext2D.drawImage() and use image decoders implemented in JavaScript to extract raw image data that was not tainted by the browser's color correction behavior.''
-
-An aspect of current implementations that is interoperable is that colors match between CSS/HTML and canvases:
-* A color value used as a canvas drawing style will have the same appearance as if the same color value were used as a CSS style
-* An image resource drawn to a canvas element will have the same appearance as if it were displayed as the replaced content of an HTML element or used as a CSS style value.
-
-This color matching behavior needs to be preserved to avoid breaking pre-existing content.
-
-Some implementations convert images drawn to canvases to the sRGB color space. This has the advantage of making the color correction behavior device independent, but it clamps the gamuts of the rendered content to the sRGB gamut, which is significantly narrower than the gamuts of some current consumer devices.
+* The color space of canvases is undefined in the current specification, although is de facto sRGB.
+* The bit-depth of canvases is likewise undefined, although is de facto 8 bits per component.
+* Images retrieved from a canvas are in an undefined color space and no color space information is encoded by ``toDataURL`` or ``toBlob``.
 
 ### Requests for this Feature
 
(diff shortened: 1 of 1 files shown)


CanvasColorSpaceProposal.md
```diff
@@ -103,61 +80,48 @@ canvas.getContext('2d', { colorSpace: "rec2020",
...
 
-#### Selecting the best encoding for the user agent's display device
-Selection should be based on the best color space match (see above). For srgb, at least 8 bits per component is recommended; for p3, 10 bits; and for rec2020, 12 bits.  The float16 format is suitable for any colorspace.  There may soon be a proposal to add a way of detecting HDR displays, possibly something like "window.screen.isHDR()" (TBD), which would be a good hint to use the float16 format.
+#### WebGL behavior
 
-#### Non-standard color spaces
-For future consideration: support could be added for color space defined using the [CSS @color-profile rule](https://www.w3.org/TR/css-color-4/#at-profile).
+* Values stored in WebGL backbuffers are in the canvas's color space.
+* Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
+* For the color encoding of ``"unorm-srgb"``, the interpretation of values written by ``gl_FragColor`` is as follows:
```
kenrussell, line 110: unorm-srgb should be unorm8-srgb.

CanvasColorSpaceProposal.md
```diff
@@ -1,52 +1,29 @@
...
+* The color space of canvases is undefined in the current specification, although is de facto sRGB.
+* The bit-depth of canvases is likewise undefined, although is de facto 8 bits per component.
+* Images retrieved from a canvas are in an undefined color space and no color space information is encoded by ``toDataURL`` or ``toBlob``.
 
 ### Requests for this Feature
 
 * <cite>[https://github.com/whatwg/html/issues/299]</cite> <blockquote><p>Allow 2dcontexts to use deeper color buffers</p></blockquote>
 * <cite>[https://bugs.chromium.org/p/chromium/issues/detail?id=425935]</cite> <blockquote><p>Wrong color profile with 2D canvas</p></blockquote>
-* Engineers from the Google Photos, Maps and Sheets teams have expressed a desire for canvases to become color managed.  Particularly for the use case of resizing an image, using a canvas, prior to uploading it to the server, to save bandwidth. The problem is that the images retrieved from a canvas are in an undefined color space and no color space information is encoded by toDataURL or toBlob.
```
kenrussell, line 29: It's unfortunate to lose this bit of feedback; I think it should be left in.

CanvasColorSpaceProposal.md
```diff
@@ -1,52 +1,29 @@
 # Color managing canvas contents
 
 ## Use Case Description
-* Contents displayed through a canvas element should be color managed in order to minimize differences in appearance across browsers and display devices. Improving color fidelity matters a lot for artistic uses (e.g. photo and paint apps) and for e-commerce (product presentation).
```
kenrussell, line 4: The mention of photo / paint / e-commerce apps seems like good motivation for the use cases to me.

kenrussell (approved):
Nice cleanups overall @ccameron-chromium . A few comments, but looks good to me.

CanvasColorSpaceProposal.md
ccameron-chromium: Re-added.

CanvasColorSpaceProposal.md
ccameron-chromium: Re-added.

CanvasColorSpaceProposal.md
ccameron-chromium: Fixed.

CanvasColorSpaceProposal.md
kenrussell: Thanks.

kenrussell (approved):
Thanks, looks good to me!

CanvasColorSpaceProposal.md
```diff
@@ -1,52 +1,33 @@
...
-Color spaces are generally tuples of:
-* Chromaticity coordinates (X,Y in CIE 1931 space)
-  * Primaries (coords for 1.0 for each of R/G/B)
-  * White-point (coord for white)
-* Transfer function (often "gamma")
 
-Within the same chromaticity, 0.0 will always be min-brightness and 1.0 is max-brightness.
-However, because of the transfer function, 0.5 will not be the average physical brightness of 0.0 and 1.0.
-You can see this by making stripes of 1.0 and 0.0, and matching that pattern's brightness to a solid grey background.
-(On my machine, striped 0/1/0/1 appears equally bright as #ACACAC, or 0.67)
```
jdashg, line 42: This is really useful background that I would like to remain visible here, but I'm happy to collaborate on better wording.

CanvasColorSpaceProposal.md
```diff
@@ -103,61 +84,48 @@ canvas.getContext('2d', { colorSpace: "rec2020",
...
 
-#### Non-standard color spaces
-For future consideration: support could be added for color space defined using the [CSS @color-profile rule](https://www.w3.org/TR/css-color-4/#at-profile).
+* Values stored in WebGL backbuffers are in the canvas's color space.
+* Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
+* For the color encoding of ``"unorm-srgb8"``, the interpretation of values written by ``gl_FragColor`` is as follows:
+  * The color value is in a linear color space.
+  * The inverse sRGB transfer function is applied only after blending has occurred.
+  * E.g, if the value ``0.5`` is written in ``gl_FragColor``, the result will be stored as ``0xbc``.
+* For all other color encodings, the value written by ``gl_FragColor`` is in the canvas' color space.
```
jdashg, line 118: This is sort of subtly wrong, or maybe just misleading: For unorm8-srgb (as implemented in hardware), 0xbc *is* the encoding of 0.5, just as unorm8's is 0x80 and float32's is 0x3f000000.
There's more ways than just gl_FragColor to set color data in WebGL, and leaning heavily on "encoding" as our word of choice communicates this with proper connotation.

I think having a description of the unusual unorm8-srgb encoding is essential, it would help to have it next to a description of the widely-understood unorm8 encoding. Having these in sequence would be helpful by drawing important parallels between the two similar-but-different encodings, jump-starting strong intuition.

```
function encodeUnorm8(val) { return val / 0xff; }
function encodeUnorm8Srgb(val) {
  if (val < 0.0031308) return 12.92 * val;
  return 1.055 * Math.pow(val, 0.41666) - 0.055;
}
```
(from https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf p182)

CanvasColorSpaceProposal.md
```diff
@@ -103,61 +84,48 @@ canvas.getContext('2d', { colorSpace: "rec2020",
...
 
-#### Selecting the best encoding for the user agent's display device
-Selection should be based on the best color space match (see above). For srgb, at least 8 bits per component is recommended; for p3, 10 bits; and for rec2020, 12 bits.  The float16 format is suitable for any colorspace.  There may soon be a proposal to add a way of detecting HDR displays, possibly something like "window.screen.isHDR()" (TBD), which would be a good hint to use the float16 format.
+#### WebGL behavior
 
-#### Non-standard color spaces
-For future consideration: support could be added for color space defined using the [CSS @color-profile rule](https://www.w3.org/TR/css-color-4/#at-profile).
+* Values stored in WebGL backbuffers are in the canvas's color space.
+* Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
+* For the color encoding of ``"unorm-srgb8"``, the interpretation of values written by ``gl_FragColor`` is as follows:
```
jdashg, line 114: s/unorm-srgb8/unorm8-srgb/

CanvasColorSpaceProposal.md
```diff
@@ -103,61 +84,48 @@ canvas.getContext('2d', { colorSpace: "rec2020",
...
+* For all other color encodings, the value written by ``gl_FragColor`` is in the canvas' color space.
 
 #### Compositing the canvas element
-Canvas contents are composited in accordance with the canvas element's style (e.g. CSS compositing and blending rules). The necessary compositing operations must be performed in an intermediate colorspace, the compositing space, that is implementation specific. The compositing space must have sufficient precision and a sufficiently wide gamut to guarantee no undue loss of precision or gamut clipping in bringing the canvas's contents to the display. Implementations should not expose color spaces that are unreasonble for the display.
+
+Canvas contents are composited in accordance with the canvas element's style (e.g. CSS compositing and blending rules). The necessary compositing operations must be performed in an intermediate colorspace, the compositing space, that is implementation specific. The compositing space must have sufficient precision and a sufficiently wide gamut to guarantee no undue loss of precision or gamut clipping in bringing the canvas's contents to the display.
```
jdashg, line 122: This is great I think, though removing "Implementations should not expose color spaces that are unreasonble for the display" is a big change!

jdashg (changes requested):
This is overall great progress!

However while I'm super happy to both have cruft cleaned out *and* to make changes to the proposal, we should be careful about the latter being mixed in as incidental changes during the former.
Please either only "remove older details" here, or repurpose this to explicitly make material changes.

CanvasColorSpaceProposal.md
brianosman: Agree that leaning on the idea of (HW) sRGB being an encoding is the better mental model. Also, I'm guessing you meant to say that encodeUnorm8 should be { return val ***** 0xff; }

ccameron-chromium: I think that support for "uint8-srgb" should be a separate WebGL spec that, in effect, makes the backbuffer be SRGB8_ALPHA8 (or similar).

I am of the opinion that it does not belong in this specification -- it doesn't have a clear meaning for 2D canvas, it's independent of color space, and it is only an option for unorm8 canvases.

CanvasColorSpaceProposal.md
ccameron-chromium: Removing "Implementations should not expose color spaces that are unreasonble for the display" is an example of the old cruft in the spec that should be removed (the original authors are long gone, and the context in which they wrote this long forgotten).

My argument for this is: A user should be able to author and manipulate content that current monitor happens to support. An example is a web app being used to crop/rotate photos using 2D canvas. If a P3 image is uploaded, then it should be manipulatable without crushing it to sRGB, regardless of the user's monitor capabilities.

jdashg: I think I support this change, but I would prefer to see it in a separate non-cleanup PR, if you can!

jdashg: @RafaelCintron is interested in how/if we handle monitor-non-native spaces, I think.

CanvasColorSpaceProposal.md
jdashg: We should probably have a larger discussion about this. I think there is desire unorm8-srgb canvas2d (@kenrussell?), but there are real implementability/feasibility concerns. Can you file an issue?

kenrussell: I don't remember off the top of my head - might have been discussed a while ago with @fserb and others. For 2D canvas photo editing applications, I can imagine it might be desirable to use the unorm8-srgb encoding to get more detail in darker regions of images.

Could we please leave the unorm8-srgb description in the WebGL-specific section of this document for the time being? Chris, would you be willing to add Jeff's code snippets?

CanvasColorSpaceProposal.md
ccameron-chromium: Left it here for now, and I'll follow up to remove that sentence and add motivation.

CanvasColorSpaceProposal.md
ccameron-chromium: Fixed -- thanks.

CanvasColorSpaceProposal.md
```diff
@@ -1,52 +1,33 @@
...
+* The [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4) specification defines several color spaces that will be used in this specification. At time of writing this specification is partially implemented on some browsers.
 
 ## Proposed Solution
 
-* Clearly define the color space of canvases.  Most current browser implementations are either color-managed or are currently actively working on becoming color-managed.  Therefore, it will be possible to specify that the default color space of canvases as 'srgb' instead of leaving it undefined in the spec (as is currently the case due to lack of interoperability).
-* Add a canvas context creation attribute to specify a color space.
-* Add a canvas context creation attribute to specify an encoding format for storing pixel values.
-* Color space and encoding format parameters are also extended to image storage interfaces that interact with canvases, such as CanvasPattern, ImageData and ImageBitmap.
+* Clearly define the default color space of canvases to be ``srgb``, and the default encoding to be 8 bits per pixel.
+* Add canvas context creation attributes to change the color space and encoding of a canvas.
```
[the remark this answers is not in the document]
ccameron-chromium: Good point -- fixed.

CanvasColorSpaceProposal.md
ccameron-chromium: I agree that some background will be helpful, but I'd prefer to leave it out for now, because (1) it's redundant with the CSS Color Level 4 spec, which specifies the color spaces and (2) I'd like to get behavior and API changes nailed down. We can pull this out of the revision history as a starting point when we get there.

CanvasColorSpaceProposal.md
ccameron-chromium: The actual bits-stored-in-the-framebuffer have identical meanings when using uint8-srgb and uint8. They are both in the sRGB color space.

What is different is that when a uint8-srgb texture is sampled in WebGL, the values are converted into a linear-sRGB color space by the texturing unit. Similarly, when values are written to the framebuffer, they are converted from linear-sRGB color space down to sRGB color space before being written (but after blending).

It's a really texture-unit-and-framebuffer-specific thing, which is why it doesn't clear meaning in other contexts (e.g, 2D canvas). A few years ago we considered exposing a 2D canvas that was sort-of-in-linear-space-but-stored-as-sRGB, and it was an unintuitive mess (and was ultimately one of the things that caused this spec's implementation to stall out as it did).

kenrussell: For the next incremental step of this specification, @jdashg and I are in agreement that we'd like the unorm8-srgb option to remain. The discussion of applicability to 2D canvas can be deferred, and would perhaps be better done on a conference call, since there's disagreement about its utility.

Can we please resolve this particular conversation? May I suggest incorporating this text:

- The encodings for a specific color value differ between <code>unorm8</code> and <code>unorm8-srgb</code> as follows:

<pre>
function encodeUnorm8(val) { return val * 0xff; }
function encodeUnorm8Srgb(val) {
  if (val < 0.0031308) return 12.92 * val;
  return 1.055 * Math.pow(val, 0.41666) - 0.055;
}
</pre>

jdashg: Oops encodeUnorm8Srgb needs to include the `* 0xff` on its returns, sorry!

ccameron-chromium: I've updated the text to include kbr's suggestion about encoding.

I'd like to continue iterating on whether GL color encoding should be specified independently of the storage format, but we can defer that discussion to some other time.

kenrussell: Thanks @ccameron-chromium . Let me call this conversation resolved, and we can open a new one against the most recent version of this pull request if necessary.

CanvasColorSpaceProposal.md
```diff
@@ -103,61 +84,56 @@ canvas.getContext('2d', { colorSpace: "rec2020",
...
+  * This means that the same content, rendered with two different canvas backing stores, may be slightly different.
+* The functions ``toDataURL`` and ``toBlob`` should produce resources that best match the fidelity of the underlying canvas.
+  * Subject to the limitations of the implementation and the requested format.
 
+#### WebGL behavior
 
-#### Selecting the best color space match for the user agent's display device
+* Values stored in WebGL backbuffers are in the canvas's color space.
+* Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
+* The encodings for specific color value differ between ``"unorm8"`` and ``"unorm8-srgb"``.
```
kenrussell, line 114: value -> values

kenrussell (approved):
Still looks good to me - one small correction.

CanvasColorSpaceProposal.md
ccameron-chromium: Oops -- updated.

kenrussell: Thanks!

ccameron-chromium:
Thanks all! Any further requests?

jdashg (approved):
Great, thanks so much!

PR #36, opened 2020-11-02
Indicate implementations should not limit exposed color spaces

ccameron-chromium:
Remove "Implementations should not expose color spaces that are unreasonble for the display." Add "Implementations should not limit the set of exposed color spaces based on the capabilities of the display."

The color gamut media queries exist to make the determination about which color space will max-out the fidelity of a given monitor.

Limiting the options for which color space can be used constrains applications in an unhelpful way. For example, consider an application to do photo cropping and zooming using a canvas or WebGL. Such an application will be using the canvas or WebGL buffer as its working space. It will want that working space to be as close a match as possible to the input image, not the monitor that the application happens to be running on.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -16,8 +16,6 @@
 * <cite>[https://bugs.chromium.org/p/chromium/issues/detail?id=425935]</cite> <blockquote><p>Wrong color profile with 2D canvas</p></blockquote>
 * Engineers from the Google Photos, Maps and Sheets teams have expressed a desire for canvases to be color managed. Particularly for the use case of resizing an image, using a canvas, prior to uploading it to the server, to save bandwidth. The problem is that the images retrieved from a canvas are in an undefined color space and no color space information is encoded by toDataURL or toBlob.
 
-
-
 ### Related Specifications
 * The [CSS Media Queries Level 5](https://www.w3.org/TR/mediaqueries-5/#color-gamut) specification defines the ``color-gamut`` media query to determine the capabilities of the current display device.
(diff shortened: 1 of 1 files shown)


CanvasColorSpaceProposal.md
```diff
@@ -89,6 +87,7 @@ canvas.getContext('2d', { colorSpace: "rec2020",
 The ``colorSpace`` attribute specifies the color space for the backing storage of the canvas.
 * Color spaces match their respective counterparts as defined in the [predefined color spaces](https://www.w3.org/TR/css-color-4/#predefined) section of the [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4) specification.
 * When an unsupported color space is requested, the color space shall fall back to ``"srgb"``.
+* Implementations should not limit the set of exposed color spaces based on the capabilities of the display.
```
kenrussell, line 90: The pull request comment "The color gamut media queries exist to make the determination about which color space will max-out the fidelity of a given monitor." is directly related to this statement. Could something similar be added as a second sentence of this bullet point, or perhaps made into non-normative text?

kenrussell (commented):
What's the consequence of this change? Could it mean that applications might choose an extended range color space that the monitor can't natively display, and that compositing

CanvasColorSpaceProposal.md
ccameron-chromium: Good point -- added a sentence to the bullet point clarifying this.

kenrussell (approved):
Very good, thanks. Looks good to me!

jdashg (approved):

jdashg:
Would like feedback from @RafaelCintron and @grorg

grorg (approved):

RafaelCintron (approved):
LGTM.

kenrussell:
Thanks all for your reviews. Merging now.

PR #37, opened 2020-11-16
Separate canvas storageType and colorEncoding

ccameron-chromium:
Separate storageType and colorEncoding canvas creation attributes. This clarifies that the color values expressible by a canvas is unaffected by its color encoding function (rather, only its color space and storage type), and mirrors other OpenGL specifications.

The OpenGL specifications have a FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING property that indicates the color encoding function that is applied to values to be written to the framebuffer. This can take on the values of LINEAR and SRGB, depending on the the function that is applied.

Mirror this behavior by separating out a "colorEncoding" from the "storageType" attribute.

The name "storageType" is used to mirror the attributes used by ImageData. There is no attachment to this particular name, only to the idea that it be the same as ImageData.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -23,7 +23,7 @@
 ## Proposed Solution
 
 * Clearly define the default color space of canvases to be ``srgb``, and the default encoding to be 8 bits per pixel.
-* Add canvas context creation attributes to specify the color space and encoding of a canvas.
+* Add canvas context creation attributes to specify the color space, storage format, and color encoding function of a canvas.
 * Clarify behavior for drawing into canvases, compositing canvases, and exporting canvas contents.
 * Add mechanisms to specify the color space and encoding of ``ImageData`` objects.
 
@@ -39,9 +39,9 @@ enum CanvasColorSpaceEnum {
   "rec-2020",
 };
 
-enum CanvasColorEncodingEnum {
-  "unorm8",      // default, 0.5 encoded as 0x80
-  "unorm8-srgb", // 0.5 encoded as 0xbc
+enum CanvasStorageFormatEnum {
+  "unorm8",      // default
+  "unorm8-srgb", // Same as unorm8, but with an sRGB color encoding function
   "float16",     // IEEE 754
 };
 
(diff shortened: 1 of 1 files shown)


brianosman (commented):
I understand the motivation here, but on Skia we abandoned this direction, and made sRGB-encoded bytes be a storage format. With this change, { storageType = "float16", colorEncoding = "srgb" } looks legal, but doesn't make sense (it doesn't map to WebGL, or any GPU).

From the rest of the current draft, the idea is that the encoding will only be relevant when using WebGL, not for canvas2D operations (ie, a GPU-backed implementation will need to use something like EXT_sRGB_write_control)?

brianosman:
... rephrasing my thoughts: There should only be one variable, not two -- they're not orthogonal. That single variable describes how values are stored, which encompasses how many bits, and what the interpretation of those bits are. In all modern GPU APIs, there is no attempt to decouple the concepts, either: VK_FORMAT, MtlPixelFormat, and DXGI_FORMAT are all enums with specific _SRGB entries for the particular formats that use that encoding function. Of course, this also suggests that a name involving "format" would be appropriate.

kenrussell:
I agree with @brianosman 's assessment, and would prefer to leave the spec as-is. It would be good to implement the current spec and work with Unity in particular to make sure that it eliminates the full-screen blit at the end of their linear-gamma rendering pipeline.

@jdashg what do you think?

ccameron-chromium:
I agree with @brianosman's comments as well, namely, we should call this "storageFormat" or "pixelFormat", not "colorEncoding". In the subsection about WebGL, we should add a note about the GL color encoding. I'll update the wording here when I get a chance.

ccameron-chromium:
I've updated the patch to clarify the naming (going with "storageFormat" instead of "colorEncoding") and to use the precise formulas from the GLES spec (which aren't 8-bit  -- I initially wanted to clarify the rounding behavior, then I found that the GLES spec is fully floating-point).

kenrussell (approved):
Looks good to me.

In a follow-on, do you think the CanvasStorageFormatEnum and ImageDataStorageType enums could be reconciled or combined?

ccameron-chromium:
Thanks!

> In a follow-on, do you think the CanvasStorageFormatEnum and ImageDataStorageType enums could be reconciled or combined?

I'm not sure which is more appropriate -- the accepted values for the two are different -- (unorm8, unorm8-srgb, float16) vs (uint8, uint16, float32).

We should probably sit down together to figure out these final details.

PR #39, opened 2020-12-02
Clarify extended values and high dynamic range behavior

ccameron-chromium:
Clarify extended values and high dynamic range behavior

Address the outstanding issue of how to handle floating point pixel values outside of the range of [0, 1]. Define these color values in the usual way (similar to, e.g, kCGColorSpaceExtendedSRGB and DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709 on macOS and Windows respectively).

Indicate that values outside of [0, 1] always extend the gamut beyond the canvas' color space's primaries (there is no artifical clamping).

Add a highDynamicRange creation attribute that controls whether or not values outside of [0, 1] can extend the luminance of the canvas.

Leave as an open issue how HDR metadata is to be specified (it is likely desirable for it to be dynamic) for now.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -64,11 +64,13 @@ interface CanvasStorageFormat {
 partial dictionary CanvasRenderingContext2DSettings {
   CanvasColorSpaceEnum colorSpace = "srgb";
   CanvasStorageFormatEnum storageFormat = "unorm8";
+  boolean highDynamicRange = false;
 };
 
 partial dictionary WebGLContextAttributes {
   CanvasColorSpaceEnum colorSpace = "srgb";
   CanvasStorageFormatEnum storageFormat = "unorm8";
+  boolean highDynamicRange = false;
 };
 
 partial interface CanvasRenderingContext2D {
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
This clarifies the meaning of pixel values outside of the range of [0, 1].

I added the highDynamicRange property because I feel it is import to clarify under what circumstances color values aren't to be clamped. Let me know if you'd prefer to remove this property for now, and just specify that "luminance will be clamped, but chrominance won't, unless otherwise specified"

CanvasColorSpaceProposal.md
```diff
@@ -96,6 +98,32 @@ The ``storageFormat`` attribute specifies the format for storing individual pixe
 * Support for ``"unorm8"`` is mandatory. All other formats are optional.
 * When an unsupported format is requested, the format shall fall back to ``"unorm8"``.
 
+#### The ``highDynamicRange`` canvas creation attribute
+
+The ``highDynamicRange`` attribute specifies whether or not this canvas can display luminance greater than that of the color ``(1,1,1)``.
+* It is possible to represent colors of arbitrarily high luminance using floating-point storage formats.
+* Unless ``highDynamicRange`` is set to ``true``, colors may have their luminance clamped to the standard dynamic range.
+* The motivation for this behavior is that enabling the display of high dynamic range content can come at a significant performance and power cost which should not be imposed unless explicitly requested.
```
kenrussell, line 106: Does it make sense to request:
`{ storageFormat: "float16", highDynamicRange: false }`
during canvas context creation?

I'd somewhat assume that a request of the float16 storage format would be the explicit signal that the user wants HDR display, at the performance and power cost.

Is this more intended to be a querying mechanism for whether the canvas can support HDR rendering? If so, do any existing media queries cover this for the display?

kenrussell (commented):
Sorry for the delay reviewing.

The specification of handling of values out of the [0,1] range looks good to me (without having thought through the implications for the various color spaces like display-p3 and rec-2020). My only question is about the new highDynamicRange property.

CanvasColorSpaceProposal.md
ccameron-chromium: That combination (non-HDR float16) would make sense. Values outside of the range of [0, 1] can specify colors outside of the specified gamut. An example application would be a WebGL application that is written around sRGB primaries, but wants to allow displaying images with colors outside of the sRGB gamut. These would be representable with values outside of the range of [0, 1].

There does not yet exist a media query for HDR capabilities. There really should (and maybe there is an in-development spec that I'm unaware of). YouTube is checking the bit depth of the display to determine HDR capability (which is far from ideal).

kenrussell: Thanks for explaining. On non-HDR displays I'm concerned about the statement "colors may have their luminance clamped". Will this lead to undefined behaviors if the browser doesn't do this clamping inherently, and the application renders values outside the [0.0, 1.0] luminance range?

It would be unfortunate if there were a major performance hit for `{ storageFormat: "float16", highDynamicRange: false }` because on non-HDR displays the browser had to manually composite the user's rendered canvas, clamping each channel to 1.0 in a fragment shader. Similar performance problems have been encountered with WebGL's `{premultipliedAlpha:false}` context creation parameter - it's necessary to do a blit per frame in order to get the texture in a form (premultiplied alpha) that the native OS's overlay mechanisms understand.

ccameron-chromium: I can change it to "colors *will* have their luminance clamped" -- that would be more clear.

On macOS, both modes run in full "overlay" mode.
On Windows, I think DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709 could in principle be an overlay, but only for HDR mode (and maybe not, depending on how the SDR slider implementation affects this spec).

kenrussell: Using "must" is better from a testability standpoint. I still have a slight concern that supporting the combination `{ storageFormat: "float16", highDynamicRange: false }` will impose a performance cost on some platforms, but let's try prototyping this and see how it looks on various OSs.

kenrussell (approved):
Looks good at this point! Since no other comments recently, merging now.

PR #41, opened 2021-01-07
Add unorm16 pixel format

ccameron-chromium:
Add a unorm16 pixel format.

This matches the pixel formats added in EXT_texture_norm16.


ccameron-chromium:
Small change that I felt belonged in the spec, based on our discussion of EXT_texture_norm16 earlier.

kenrussell (approved):
Looks great!

@shrekshao might be interested in this.

PR #45, opened 2021-02-23
Remove highDynamicRange canvas setting

ccameron-chromium:
Remove highDynamicRange setting from CanvasRenderingContext2DSettings and WebGLContextAttributes.

Enabling/disabling HDR will be done on the HTMLCanvasElement. Add language indicating this when discussing value clamping.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -66,13 +66,11 @@ interface CanvasStorageFormat {
 partial dictionary CanvasRenderingContext2DSettings {
   CanvasColorSpaceEnum colorSpace = "srgb";
   CanvasStorageFormatEnum storageFormat = "unorm8";
-  boolean highDynamicRange = false;
 };
 
 partial dictionary WebGLContextAttributes {
   CanvasColorSpaceEnum colorSpace = "srgb";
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
This largely-reverts the HDR parameter added earlier (it's not a complete revert, because it keeps the language related to value clamping).

Based on the W3C calls and related discussions, the HTMLCanvasElement is the right place to enable or disable HDR, and to specify HDR metadata. I'm leaving the addition of that API to a separate pull, so it can be scrutinized on its own.

jdashg (changes requested):
I'm pretty confident hdr settings should be in the same struct as colorspace settings.

jdashg:
TBH <canvas> is just a div/iframe for generated graphics, similar to an  or <video>, where its source (rendering context) is really what determines its stats, not really the element itself normally.
I'm not sure what the plan is for images/videos for hdr but I would be surprised if the expected path is to both have to set the src and also tweak other params on the elements themselves. I would expect it to be as easy as `src="hdr.webm"`!

kenrussell (approved):
LGTM - would like to see where the HDR control is added to the canvas spec.

ccameron-chromium:
Merging this, based on our discussion on Tuesday.

PR #46, opened 2021-02-23
ImageDataStorageType renaming

ccameron-chromium:
Rename ImageDataStorageType to ImageDataStorageFormat. This matches corresponding CanvasStorageFormat.

Also change uint8 and uint16 to unorm8 and unorm16. This matches CanvasStorageFormat names, and is also a better choice, because it is more descriptive, and doesn't preclude other uses of Uint16Array (e.g, for actual-unsigned-integer values, or for R5G6B5, or half-float).

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -178,14 +178,14 @@ ImageBitmap objects (unless created with ``colorSpaceConversion="none"``) should
 
 Add the following types to be used by `ImageData`.
 <pre>
-enum ImageDataStorageType {
-  "uint8", // default
-  "uint16",
+enum ImageDataStorageFormat {
+  "unorm8", // default
+  "unorm16",
   "float32",
 };
 dictionary ImageDataSettings {
   CanvasColorSpaceEnum colorSpace = "srgb";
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
Having discussed this some more, the names "unorm8" and "unorm16" are definitely better than "uint8" and "uint16". And they match the canvas options. Also, having ImageDataStorageFormat align with CanvasStorageFormat is less surprising for users.

jdashg (approved):
`Format` is ok though `Encoding` might be better? This is fine to take though.

kenrussell (approved):
LGTM

ccameron-chromium:
> `Format` is ok though `Encoding` might be better? This is fine to take though.

The term "format" is the more common term for this across the modern graphics APIs (e.g, VkFormat, DXGI_FORMAT, MTLPixelFormat), and also in OpenGL (the format and internal_format parameters).

The term "encoding" is used in the context of GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING, which is a different, graphics-pipline-specific concept that we'll want to keep distinct.

ccameron-chromium:
Thanks all!

PR #48, opened 2021-02-23
Remove outdated references to getContextAttributes

ccameron-chromium:
When this proposal was first written, the interface CanvasRenderingContext2D did not yet have a getContextAttributes method. This method now exists, and does not need to be introduced by this spec.


jdashg (approved):

jdashg:
Thanks!

jdashg:
Oops, sorry @fserb 
I'm pretty sure we want this but we can revert if you have reservations.
(This is what I get for skimming emails too quickly :))

kenrussell (commented):
LGTM

fserb:
Lgtm. :)

PR #49, opened 2021-02-23
Separate out framebufferColorEncodingSRGB

ccameron-chromium:
Add the ability to specify the use of sRGB color encoding in WebGLContextAttributes. Remove "unorm8-srgb" as a CanvasStorageFormat.

This is motivated by the fact that "unorm8-srgb" specifies a color encoding function (specific to writing to the backbuffer through shaders), but not a storage format. The interpretation of the actual bits inside the backbuffer is unaffected by the color encoding function (all access to the backbuffer via pixel paths, which, in WebGL, is limited to ReadPixels are therefore unaffected).

This is also motivated by the fact that "unorm8-srgb" is not an appropriate format to specify to 2D canvas.

Update the language in the WebGL section to clarify this.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -41,7 +41,6 @@ enum CanvasColorSpaceEnum {
 
 enum CanvasStorageFormatEnum {
   "unorm8",      // default
-  "unorm8-srgb", // Same as unorm8, but with an sRGB color encoding function
   "unorm16",
   "float16",     // IEEE 754
 };
@@ -56,7 +55,6 @@ interface CanvasColorSpace {
 
 interface CanvasStorageFormat {
   const CanvasStorageFormatEnum unorm8 = "unorm8";
-  const CanvasStorageFormatEnum unorm8Srgb = "unorm8-srgb";
   const CanvasStorageFormatEnum unorm16 = "unorm16";
   const CanvasStorageFormatEnum float16 = "float16";
 };
@@ -71,6 +69,9 @@ partial dictionary CanvasRenderingContext2DSettings {
 partial dictionary WebGLContextAttributes {
   CanvasColorSpaceEnum colorSpace = "srgb";
   CanvasStorageFormatEnum storageFormat = "unorm8";
+  boolean framebufferColorEncodingSRGB = false;
+};
+
 };
 </pre>
 
@@ -126,14 +127,23 @@ else
 
 #### WebGL behavior
 
-* Values stored in WebGL backbuffers are in the canvas' color space.
-* Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
-* The color encoding function is defined as follows:
-* For formats of ``"unorm8"``, ``"unorm16"``, and ``"float16"`` the color encoding function is:
+Values stored in WebGL's default back buffer are in the canvas' color space.
+This includes values accessed directly (e.g, through `ReadPixels`) and all values written via shaders in the graphics pipeline.
+
+The context attribute ``framebufferColorEncodingSRGB`` controls the color encoding function applied to the fragment shader's color output value when writing to the WebGL default back buffer.
+* If set to ``false`` then:
+  * The behavior is the same as when ``FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING`` _is not_ ``SRGB`` (as described in Section 4.1.8 "sRGB Conversion" of the [OpenGL ES 3.0.6 specification](https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf)).
+  * The color encoding function is the identity function.
+  * Consequently, the value assigned to the fragment shader's color output variable is interpreted as being in the canvas' color space.
+  * In code, the precise color encoding function is:
 <pre>
 function encode(val) { return val; }
 </pre>
-* For the format ``"unorm8-srgb"``, the color encoding function is:
+* If set to ``true`` then:
+  * The behavior is the same as when ``FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING`` _is_ ``SRGB`` (as described in Section 4.1.8 "sRGB Conversion" of the [OpenGL ES 3.0.6 specification](https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf)).
+  * The color encoding function is the sRGB encoding transfer function, which converts from linear values to sRGB values.
+  * Consequently, the value assigned to the fragment shader's color output variable can be interpreted as being in a linear version of the canvas' color space. (This interpretation applies only to the two color spaces that use the sRGB transfer function, namely, ``"srgb"`` and ``"display-p3"``. It does not apply to ``"rec2020"``, because that space has a non-sRGB transfer function).
+  * In code, the precise color encoding function is:
 <pre>
 function encode(val) {
   if (val < 0) return 0.0;
@@ -142,7 +152,8 @@ function encode(val) {
   else return 1.0;
 }
 </pre>
-* For all formats, the values stored in the framebuffer are in the canvas' color space.
+
+The ``WebGLContextAttributes`` settings of ``alpha:false`` and ``framebufferColorEncodingSRGB:true`` are only compatible with ``storageFormat:"unorm8"``. If any other value of ``storageFormat`` is specified, then ``alpha`` will be overridden to ``true`` and ``framebufferColorEncodingSRGB`` will be overridden to ``false``.
 
 #### Compositing the canvas element
 


ccameron-chromium:
Putting this up for discussion in our meeting today. The more time I spend thinking about this, the more I'm convinced that this is the right framing for the feature, given how this plays out with 2D (where it isn't applicable) and WebGPU (where the actual pixel formats are exposed directly).

It's fairly isolated, and could potentially be ratified ahead of the rest of the color specification.

CanvasColorSpaceProposal.md
```diff
@@ -72,6 +70,7 @@ partial dictionary CanvasRenderingContext2DSettings {
 partial dictionary WebGLContextAttributes {
   CanvasColorSpaceEnum colorSpace = "srgb";
   CanvasStorageFormatEnum storageFormat = "unorm8";
+  boolean srgbColorEncoding = false;
```
jdashg, line 73: I would rather say "unorm8Srgb cannot be used with canvas2d".
We would need to forbid srgbColorEncoding with any format other than unorm8 otherwise, which is the same class of restriction.

jdashg (changes requested):

CanvasColorSpaceProposal.md
kainino0x, line 73: Probably unnecessary idea, but (if this is accepted) perhaps this could be expressed more explicitly with an enum involving "perceptually-linear" and "physically-linear"? I'm having trouble figuring out if this makes any sense though so feel free to ignore.

ccameron-chromium: This feature does not belong in this spec, and is only a source of confusion. It affects only WebGL, and should be exposed in a way that centers WebGL. It is meaningless for 2D canvas, and is not needed for WebGPU.

I have updated the patch to remove it.

This feature is useful, but should be added in a separate, WebGL-centric specification.

kenrussell: Rather than split off this functionality into the WebGL spec, could we allow `unorm8` and `srgb-linear` to work with WebGL, and add a couple of caveats that readbacks via `readPixels` will yield slightly different results than expected?

The rendering and display portions should work the same as with the other rendering APIs. Most users won't encounter the behavioral difference with read-back pixel values, so it would be nice to request the canvas rendering context in the same way for WebGL as with e.g. 2D canvas.

ccameron-chromium: I think I see the source of everyone's disagreement here. EGL's definition of "color space" was a complete screw-up, and confuses everyone who reads it. Fortunately, EGL was the only thing that screwed this up, and everything else (Vulkan, Metal, D3D) didn't screw it up. So we can just throw EGL (well, this particular part of it) under the bus and all agree on everything!!!

Let's look at [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkColorSpaceKHR.html).

- The VkSurfaceFormatKHR specifies a VkFormat and a VkColorSpaceKHR independently.
- The color space can be VK_COLOR_SPACE_SRGB_NONLINEAR_KHR or VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT. It can also be things like DISPLAY_P3_NONLINEAR, BT2020_LINEAR, HDR10_HLG, or a [bunch of other stuff](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkColorSpaceKHR.html). 
- The pixel format can be VK_FORMAT_R8G8B8A8_UNORM or VK_FORMAT_R8G8B8A8_SRGB (the later specifying that sRGB color encoding is desired). It can also be many other things like A2B10G10R10, R16G16B16A16_SFLOAT, and [many others](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkFormat.html).
- The "I want 8-bit physically based rendering" configuration is VK_COLOR_SPACE_SRGB_NONLINEAR_KHR with VK_FORMAT_R8G8B8A8_SRGB.
- Critically, the color space and the pixel format can be mixed-and-matched.

Let's look at [Metal](https://developer.apple.com/documentation/quartzcore/cametallayer?language=objc)

- The CAMetalLayer specifies a colorspace and pixelFormat separately
- Color space can be kCGColorSpaceSRGB or kCGColorSpaceLinearSRGB. It can also be AdobeRGB, DisplayP3, ITU_R2020, or another [whole bunch of stuff](https://developer.apple.com/documentation/coregraphics/cgcolorspace/color_space_names?language=objc).
- Pixel format can be MTLPixelFormatBGRA8Unorm or MTLPixelFormatBGRA8Unorm_sRGB (the latter having sRGB color encoding). It can also be the same bazillion [other formats](https://developer.apple.com/documentation/metal/mtlpixelformat?language=objc).
- The "I want 8-bit physically based rendering" configuration is kCGColorSpaceSRGB with MTLPixelFormatBGRA8Unorm_sRGB
- Critically, the color space and the pixel format can be mixed-and-matched.

Let's look at D3D.

- An IDXGISwapChain has a [DXGI_FORMAT](https://docs.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format) and can have a [DXGI_COLOR_SPACE_TYPE](https://docs.microsoft.com/en-us/windows/win32/api/dxgicommon/ne-dxgicommon-dxgi_color_space_type#see-also) set on it.
- The format can be DXGI_FORMAT_R8G8B8A8_UNORM or DXGI_FORMAT_R8G8B8A8_UNORM_SRGB, or again, many other things.
- These formats can be used with the color space DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709 (which is sRGB). There are some other options, but it's pretty restrictive.
- The "I want 8-bit physically based rendering" configuration is DXGI_FORMAT_R8G8B8A8_UNORM_SRGB with DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709
- This has the same structure as the other examples, but it doesn't allow as much mixing-and-matching. Most of the supported color spaces are for video.

Let's look at GLX and WGL.
- In 2008, [ARB_framebuffer_sRGB](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_framebuffer_sRGB.txt) was ratified.
 - It added the ability to enable/disable sRGB color encoding on the backbuffer using FRAMEBUFFER_SRGB
 - To be able to use that, one needs to specify WGL[GLX]_FRAMEBUFFER_SRGB_CAPABLE_ARB
 - This enables what people had been able to do with framebuffer objects bound to GL_SRGB8_ALPHA8 textures. The spec indicates the interactions with GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING
- GLX never added color spaces (as far as I can see).

Then EGL came along and completely screwed up.
- The original sin is in 2014 with [KHR_gl_colorspace](https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_gl_colorspace.txt).
 - This spec wanted to do what was in ARB_framebuffer_sRGB. It has the same language about GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING.
 - But it uses the word "color space" incorrectly. It allows specifying EGL_GL_COLORSPACE_KHR to be either SRGB or LINEAR. This has nothing to do with the color space of the framebuffer. It's a total misnomer.
 - Things get even worse. Consider the quote "[if set to] EGL_GL_COLORSPACE_SRGB_KHR, then [the color space is] a non-linear, perceptually uniform color space is assumed, with a corresponding GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING value of GL_SRGB". This is also up a creek because sRGB is a perceptually uniform by a long shot. Something like [CIELUV](https://en.wikipedia.org/wiki/CIELUV) is. sRGB happens to be close-ish-maybe to perceptually uniform for grayscale values.
- Then EGL tried to add actual color spaces in 2017 with [EGL_EXT_gl_colorspace_scrgb](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_gl_colorspace_scrgb.txt), [EXT_gl_colorspace_scrgb_linear](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_gl_colorspace_scrgb_linear.txt), and [EXT_gl_colorspace_display_p3](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_gl_colorspace_display_p3.txt), among [others](https://www.khronos.org/registry/EGL/extensions/EXT/).
 - These try to keep using the EGL_GL_COLORSPACE_KHR enum because it's the name for the thing that they want.
 - EGL_GL_COLORSPACE_SCRGB_EXT and EGL_GL_COLORSPACE_SCRGB_LINEAR_EXT are then intended only to be used with floating-point pixel formats.
 - I haven't managed to wrap my head around when to use P3, P3_LINEAR, and P3_PASSTHROUGH

The whole "srgb" versus "linear" thing is horribly confusing because EGL made it incredibly confusing. I honestly can't keep EGL straight in my head.

But, Vulkan, Metal, D3D, and even GLX, are very clear. So I think the goal should be to mirror the sane APIs, and their language. WebGPU fits this mix very well.

WebGL is a bit odd because we don't specify a pixel format (there's no "use GL_RGBA", there's "use alpha:true"). I think that the most consistent thing would be to have a srgbColorEncoding parameter (like the first version of this patch). It also mirrors the alpha behavior in that it can only be used with 8-bit textures (there is no RGB texture that is half-float -- you can't get rid of the alpha channel). I also think it should be in a separate specification (and I think it would be ratified faster than this spec, because it'd be so small).

(Apologies for any typos -- I have to run now, but I wanted to get this out earlier. I'll try wrapping my head around the various EGL options, so that at least I can document it here).

kenrussell: Thanks for the in-depth background Chris.

What do you propose? In https://github.com/WICG/canvas-color-space/pull/42 - now reverted, because I merged it before Jeff had a chance to review it - "linear" versions of the various color spaces were added.

Do you suggest to factor out the transfer function somehow? Change the definitions of the color spaces and pixel formats? Something else?

Basically - it would be good to have some way to say at the JavaScript level that the user would like a GL_SRGB8_ALPHA8 texture attached as WebGL's default framebuffer. Ideally, reinterpreting whatever enums are defined in this spec, e.g. for WebGPU, and not bolted on separately to the WebGL spec. A little fudge factor can be added either to this spec or WebGL's to indicate that operations like `readPixels` won't return correct values like the WebGPU API would.

ccameron-chromium: > Do you suggest to factor out the transfer function somehow? Change the definitions of the color spaces and pixel formats? Something else?

We should factor out not not the transfer function, but the backbuffer's color encoding (aka FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING of framebuffer zero).
- We have to keep very clear that "color space" is completely orthogonal to FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING
- That's the mistake that EGL made that is causing so much confusion.
- FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING being set to SRGB causes a specific mathematical function to be applied to the gl_FragColor values after the blending stage of the pipeline.
  - That mathematical function happens to be the transformation from linear-sRGB space to sRGB space. This means that gl_FragColor is, effectively, in linear-sRGB space in that state.

I advocate that we be as explicit about it as possible, by adding a parameter WebGLContextAttributes.
- It could be "framebufferColorEncoding" and be set to "srgb" or "linear" (default "linear"), if we want to be consistent with OpenGL's terminology.
  - Going with "framebufferColorEncodingFunction" as "linear-to-srgb" or "identity" would be more precise, but breaking with tradition
- Alternatively, a boolean enableSrgbFramebufferColorEncoding and set to "true" or "false" (default "false"), if we want to try to avoid the bad "linear" terminology.
  - This is the best way to go.
  - This aligns with the way WebGL uses alpha:"false".
    - That is, in effect, specifying GL_RGB
    - It also only applies to 8 bit backbuffers
  - This mirrors GLX_FRAMEBUFFER_SRGB_CAPABLE_ARB from [ARB_framebuffer_sRGB](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_framebuffer_sRGB.txt)

When "enabled", by whatever, means, the transformation described in Section 17.3.9 of the [OpenGL 4.3 spec](https://www.khronos.org/registry/OpenGL/specs/gl/glspec43.core.pdf) would be applied.

> Ideally, reinterpreting whatever enums are defined in this spec, e.g. for WebGPU, and not bolted on separately to the WebGL spec.

Having a storage format of "unorm8-srgb" can be made to work. But a separate boolean is strictly superior. Four reasons
- It matches how alpha impacts the backbuffer's GL format
- It doesn't pollute the general-purpose CanvasStorageFormat enum (which is used by 2D canvas)
- It matches ARB_framebuffer_sRGB, which got this right
- It is orthogonal to the rest of the spec, and could be fast-tracked in its own spec.

> A little fudge factor can be added either to this spec or WebGL's to indicate that operations like readPixels won't return correct values like the WebGPU API would.

Having gone through the details of how EGL screwed this up, I come down on the side of a hard no to this. There is an integrity to separating FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING from the color space. EGL tried to munge them together and ended up with a mess. Vulkan/Metal/D3D did it right.

ccameron-chromium: I've edited my long-explanation-of-EGL comment to be more comprehensive and hopefully easier to follow.

I now very much understand the revulsion to the term "linear" and "srgb" -- the original EGL specs abused them horribly. But it's just the EGL specs (and one WGL spec). Everything else, and all of the newer APIs (Vulkan/MetalD3D) use these terms in a clear and coherent way. I don't think we should let KHR_gl_colorspace poison the standard terminology for us. (Though we should probably give ample pointers to ensure the misunderstandings that resulted in that spec are not repeated).

I've updated the patch to expose what I feel is the best API, for the reasons I described. As I mentioned before, I think that this should be in a separate spec, because it **is** orthogonal functionality and should be treated as such (and not doing so has dire consequences). But, I'm content to hitch it along for the ride here, in this form, if that makes everyone else happy.

CanvasColorSpaceProposal.md
```diff
@@ -139,14 +138,23 @@ else
 
 #### WebGL behavior
 
-* Values stored in WebGL backbuffers are in the canvas' color space.
-* Values written by ``gl_FragColor`` use the primaries of the canvas' color space.
-* The color encoding function is defined as follows:
-* For formats of ``"unorm8"``, ``"unorm16"``, and ``"float16"`` the color encoding function is:
+Values stored in WebGL backbuffers are in the canvas' color space.
```
kenrussell, line 141: Correcting preexisting terminology: perhaps "WebGL's default back buffer"?

CanvasColorSpaceProposal.md
```diff
@@ -139,14 +138,23 @@ else
...
+This includes values accessed directly (e.g, through `ReadPixels`) and all values written via shaders in the graphics pipeline.
+
+The context attribute ``framebufferColorEncodingSRGB`` controls the color encoding function applied to values assigned to ``gl_FragColor`` after the blending stage of the graphics pipeline.
```
kenrussell, line 144: Let's take the opportunity to correct this terminology since ESSL 300 shaders (used by WebGL 2.0) have user-defined fragment shader output variables.

"...applied to the fragment shader's color output value..."

Not sure whether it's necessary to mention again here that this is only done when writing to WebGL's built-in back buffer, rather than user-defined framebuffers.

CanvasColorSpaceProposal.md
```diff
@@ -139,14 +138,23 @@ else
...
+* If is set to ``false`` then:
+  * The behavior is the same as when ``GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING`` is ``GL_LINEAR`` (as described in section 17.3.9 of the [OpenGL 4.3 specification](https://www.khronos.org/registry/OpenGL/specs/gl/glspec43.core.pdf)).
```
kenrussell, line 146: WebGL is based on the OpenGL ES specifications. Could you please reference section 6.1.13 "Framebuffer Object Queries" of the ES 3.0.6 spec, https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf ?

CanvasColorSpaceProposal.md
```diff
@@ -139,14 +138,23 @@ else
...
+  * The color encoding function is the identity function.
+  * Consequently, the value assigned to ``gl_FragColor`` is interpreted as being in the canvas' color space.
```
kenrussell, line 148: ...value assigned to the fragment shader's color output variable...

CanvasColorSpaceProposal.md
```diff
@@ -139,14 +138,23 @@ else
...
+  * In code, the precise color encoding function is:
 <pre>
 function encode(val) { return val; }
 </pre>
-* For the format ``"unorm8-srgb"``, the color encoding function is:
+* If set to ``true`` then:
+  * The behavior is the same as when ``GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING`` is ``GL_SRGB``  (as described in section 17.3.9 of the [OpenGL 4.3 specification](https://www.khronos.org/registry/OpenGL/specs/gl/glspec43.core.pdf)).
```
kenrussell, line 154: Please cite ES 3.0.6 spec per above.

CanvasColorSpaceProposal.md
```diff
@@ -139,14 +138,23 @@ else
...
+  * The color encoding function is the sRGB encoding transfer function, which converts from linear values to sRGB values.
+  * Consequently, the value assigned to ``gl_FragColor`` can be interpreted as being in a linear version of the canvas' color space. (This interpretation applies only to the two color spaces that use the sRGB transfer function, namely, ``"srgb"`` and ``"display-p3"``. It does not apply to ``"rec2020"``, because that space has a non-sRGB transfer function).
```
kenrussell, line 156: ...value assigned to the fragment shader's color output variable...

kenrussell (approved):
OK to move forward. My own preference would be to align this functionality with how it'll be specified for WebGPU, but as long as there's some way to get at the functionality and it's not pushed off to the WebGL spec it's OK. A few comments; please address before merging. Thanks.

kenrussell (approved):
Also - please make sure the commit message for this pull request is appropriately retitled when squashing and merging. Thanks.

CanvasColorSpaceProposal.md
ccameron-chromium: The separate boolean has the advantages that
- it mirrors how alpha impacts the GL format for the backbuffer
- it doesn't add a format that cannot be used with 2D canvas

CanvasColorSpaceProposal.md
ccameron-chromium: Fixed.

CanvasColorSpaceProposal.md
ccameron-chromium: Done. Added "when writing to the WebGL default back buffer"

CanvasColorSpaceProposal.md
ccameron-chromium: Done, but I link to section 4.1.8 "sRGB Conversion", because that's where the behavior is defined.

Also changed this from "is ``GL_LINEAR``" to "is not ``SRGB``", because that's the language of the spec.

CanvasColorSpaceProposal.md
ccameron-chromium: Done

CanvasColorSpaceProposal.md
ccameron-chromium: Done (same caveats as above)

CanvasColorSpaceProposal.md
ccameron-chromium: Done

CanvasColorSpaceProposal.md
ccameron-chromium: Done

CanvasColorSpaceProposal.md
ccameron-chromium: Done

jdashg:
The encoding is naturally intrinsic to the format across all our APIs, and I don't think splitting it out here makes things better.
If canvas2d can't use it (canvas2d probably can't feasibly use float16 either, and I'm not even sure about unorm16), just forbid it from canvas. The different (format,encoding) tuple you make here is really just (format,) in other APIs.

ccameron-chromium:
We had a discussion offline, and the conclusion is that WebGL will have a separate proposal to introduce the default back buffer pixel format. This matches WebGPU (where a pixel format already can be specified). This also allows specifying a wide range of formats (which may not be ideal to apply to 2D canvas), including RGBA8, SRGB8_ALPHA8, RGB10_A2, RGBA16_EXT, RGBA16F, and probably others.

Meanwhile, this spec will remove the concept of storageFormat, and will limit itself to DisplayP3. Introducing a color space parameter to 2D Canvas, WebGL, WebGPU, ImageData, and ImageBitmap is a bit step on its own.

We will then introduce a new proposal that adds Ultra Wide Color Gamut (Rec2020) and High Dynamic Range. This proposal will require adding >8 bit per pixel support to 2D Canvas, and we'll cross that bridge in that spec.

I'm very happy with this outcome, and I think it will result in everything being standardized and shipped sooner.

PR #51, opened 2021-02-26
Reorganize intro, emphasizing goal and solution outline

ccameron-chromium:
Update the introduction section of the proposal to start with the current limitations and goals, and end with an outline of the proposal.

Also update the wording to reflect the upcoming changes that will be made removing storage format parameters. This will be followed by patches that
- remove the storage parameters and rec2020
- break out the Proposed Solution section into sections of CanvasColorSpaceEnum, 2D Canvas, WebGL, WebGPU, HTMLCanvasElement, ImageData
- add ImageBitmap
- add worked examples
At that point, I think we can freeze things (except perhaps for defining edge case behavior).

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -1,14 +1,27 @@
 # Color managing canvas contents
 
-## Use Case Description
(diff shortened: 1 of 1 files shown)


jdashg (approved):

kenrussell (commented):
LGTM after the fact.

PR #52, opened 2021-02-26
Remove CanvasStorageFormat, ImageDataStorageFormat, and Rec2020

ccameron-chromium:
This functionality will be exposed in other proposals.

The WebGL parts of CanvasStorageFormat will go into a separate proposal.

The 2D Canvas storage format is being deferred until a later proposal that covers ultra wide color gamut and high dynamic range.

Because Rec2020 is only really useful with more than 8 bits per pixel, it is removed as well.


jdashg:
The WebGL side of this is being handled here: https://github.com/KhronosGroup/WebGL/pull/3220

jdashg (approved):

PR #53, opened 2021-02-26
Move IDLs into sections

ccameron-chromium:
In the Proposed Solution section, there exist sub-sections for each part of the API (2D Canvas, WebGL, WebGPU, HTMLCanvasElement, and ImageData). Move the IDL changes for each of those API parts into those sub-sections (this is how ImageData is already, but do the same for the other parts)

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -55,7 +55,7 @@ This proposal will also clarify:
 
 The remainder of this section discusses this functionality and behavior in detail.
 
-### Processing Model
+### Color Spaces
 
 IDL:
 <pre>
@@ -72,16 +72,6 @@ interface CanvasColorSpace {
   const CanvasColorSpaceEnum srgb = "srgb";
   const CanvasColorSpaceEnum displayP3 = "display-p3";
 };
-
-// Feature activation:
-
-partial dictionary CanvasRenderingContext2DSettings {
-  CanvasColorSpaceEnum colorSpace = "srgb";
-};
-
-partial dictionary WebGLContextAttributes {
-  CanvasColorSpaceEnum colorSpace = "srgb";
-};
 </pre>
 
 Example:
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
In this patch I intentionally just move things around, and don't change the language in any of the sections. The remaining changes to the spec are
- add the ImageBitmap section that we've discussed at W3C
- add examples
- do a final pass of language (it has the fingerprints of many authors)

ccameron-chromium:
Post-weekend ping. This change is moving text around (and intentionally not changing the text -- trying to keep the "move stuff" and "change stuff" diffs separate).

CanvasColorSpaceProposal.md
```diff
@@ -72,43 +72,47 @@ interface CanvasColorSpace {
...
 </pre>
 
 Example:
 <pre>
 canvas.getContext('2d', { colorSpace: "display-p3"} );,
 </pre>
 
-#### The ``colorSpace`` canvas creation attribute
-
 The ``colorSpace`` attribute specifies the color space for the backing storage of the canvas.
```
kenrussell, line 82: At this point in the new text, the "colorspace" attribute hasn't been defined in IDL yet. Should there be a mention that it's defined in subsequent sections?

kenrussell (approved):
Generally looks good; one thought.

CanvasColorSpaceProposal.md
ccameron-chromium: You're right.

This is a part that will need re-working. I had started doing bits of clean-up along with the restructuring, but the diffs became unwieldy.

Once this general re-structure in is in place (and once the ImageBitmap part is added), then I'm going to do a final pass of tightening up the verbiage and fixing these sorts of dependencies.

kenrussell: OK, fine by me to move forward with this and continue to do cleanups.

ccameron-chromium:
Ping on this again -- does anyone else want to review this (this re-org is blocking the changes to the verbiage of each section).

(Thanks Ken for the review!)

jdashg (approved):

jdashg:
Merged, thanks!

PR #54, opened 2021-03-10
Update WebGL color space and texture upload

ccameron-chromium:
Move WebGL's color space to be specified as WebGLRenderingContextBase attribute.

Update the language to be clear about the when this attribute is latched for compositing.

Indicate that the WebGLRenderingContextBase color space is to be used for color conversion during texture upload.

Clarify that using color conversion in this path is likely less efficient than ImageBitmap's color conversion. Update the ImageBitmap section to include the color space parameter in ImageBitmapOptions.

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -107,26 +107,22 @@ partial dictionary CanvasRenderingContext2DSettings {
 
 When an unsupported color space is requested, the color space will fall back to ``"srgb"``.
 
-#### WebGL
+### WebGL
 
 IDL:
 <pre>
-// Feature activation:
-
-partial dictionary WebGLContextAttributes {
-  PredefinedColorSpaceEnum colorSpace = "srgb";
+partial interface WebGLRenderingContextBase {
+  attribute PredefinedColorSpaceEnum colorSpace = "srgb";
 };
 </pre>
 
-Values stored in WebGL's default back buffer are in the canvas' color space.
+Values stored in WebGL's default back buffer are to be interpreted as being in the color space specified by the ``colorSpace`` attribute.
+Changing the ``colorSpace`` attribute is not destructive to the contents of the default back buffer.
 
-Pixel values accessed directly (e.g, through ReadPixels) are in the canvas' color space.
+The ``UNPACK_COLORSPACE_CONVERSION_WEBGL`` pixel storage parameter indicates the color space conversion that will be applied to ``HTMLImageElement`` sources passed to ``texImage2D`` and ``texSubImage2D``.
+In implementations in which a ``UNPACK_COLORSPACE_CONVERSION_WEBGL`` of ``BROWSER_DEFAULT_WEBGL`` causes a conversion to sRGB color space, it is recommended that this behavior be changed to be a conversion to the color space indicated by ``colorSpace``.
 
-The values written to the default backbuffer through the graphics pipeline are also in the canvas' color space.
-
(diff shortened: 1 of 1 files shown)


ccameron-chromium:
Putting this up based on our discussion. I included the ImageBitmap part as well, because it interacts with UNPACK_COLORSPACE_CONVERSION_WEBGL (I can put it in a separate PR if that's preferred)

kenrussell (approved):
Looks good to me as a step forward!

ccameron-chromium:
Were there any more comments on this?

CanvasColorSpaceProposal.md
```diff
@@ -152,7 +150,15 @@ This is in contrast with luminance, which is to be clamped to the maximum standa
...
-ImageBitmap objects (unless created with ``colorSpaceConversion="none"``) should keep track of their internal color space, and should store their contents at highest fidelity possible, subject to implementation limitations.
+IDL:
+<pre>
+partial dictionary ImageBitmapOptions {
+  CanvasColorSpaceEnum colorSpace = "srgb";
+}
+</pre>
+
+When creating an ``ImageBitmap``, if the ``colorSpaceConversion`` entry of the specified ``ImageBitmapOptions`` is not ``"none"``, then the internal color space for the resulting ``ImageBitmap`` should be the color space that is specified by the ``colorSpace`` entry of the ``ImageBitmapOptions``.
+If that ``ImageBitmap`` is then used as input to populate a WebGL or WebGPU texture, then the pixel values written to the texture should represent the ``ImageBitmap``'s source contents in the specified color space.
```
lexaknyazev, line 161: What happens when the `colorSpace` of the image bitmap does not match the `colorSpace` of the drawing context at a time of texture upload?

ccameron-chromium: Note that even if we didn't have the new UNPACK_COLORSPACE_CONVERSION_WEBGL behavior, we'd still have this problem. The default behavior (on some implementations) is to convert to sRGB, and that conversion would need to disabled somehow.

I'll clarify that "Beware that the ImageBitmap may be converted to the WebGLRenderingContext color space during texture upload unless UNPACK_COLORSPACE_CONVERSION_WEBGL is set to NONE". I'll add that to the PR in the morning.

There is an alternative, which is to say that ImageBitmaps are special and cause UNPACK_COLORSPACE_CONVERSION_WEBGL to always be treated as NONE. That feels like something that will become messy to me.

Note for another change to add: There is no such thing as a "none" color space (such a thing breaks any color management system). Internally, "none" is "replace the profile with sRGB without converting". This should be updated to indicate that "none" is "replace the profile with `colorSpace` without converting".

lexaknyazev: WebGL spec defines `UNPACK_COLORSPACE_CONVERSION_WEBGL` only for `HTMLImageElement`. It should not affect image bitmaps.

kenrussell: In WebGL when ImageBitmaps are uploaded to textures, *no* colorspace conversion, y-flip, etc. are done. This is by design. The `UNPACK_COLORSPACE_CONVERSION_WEBGL`, `UNPACK_FLIP_Y_WEBGL` and `UNPACK_PREMULTIPLY_ALPHA_WEBGL` context settings are all ignored when uploading ImageBitmaps; these settings are applied during `ImageBitmap` creation.

The same rule will apply for the new `colorSpace` setting. The context's setting will be ignored when upoading an `ImageBitmap`.

ccameron-chromium: Thanks for the clarification. I've updated the language to make this (lack of) interaction clear.

jdashg (approved):

PR #59, opened 2021-03-23
Rename CanvasColorSpace to PredefinedColorSpace

ccameron-chromium:
The CSS Color Module defines a set of [predefined color spaces](https://drafts.csswg.org/css-color/#predefined).

Change CanvasColorSpaceEnum to PredefinedColorSpaceEnum. Two reasons
- CanvasColorSpaceEnum refers to these predefined color spaces, so the name is a better fit.
- CanvasColorSpaceEnum is used in several non-canvas places (e.g, ImageData, ImageBitmap, WebGLRenderingContext), so the new name removes this awkward scoping

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -61,16 +61,16 @@ IDL:
 <pre>
 // Feature enums:
 
-enum CanvasColorSpaceEnum {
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Good idea, looks good!

jdashg (approved):

PR #60, opened 2021-04-25
References the HDR work taking place in the Color-on-the-web CG

### LINKED ISSUE #15: Question on scope + P3
> Canvases should be able to take advantage of the full color gamut and dynamic range of the display device.

There are literally thousands of different display device characteristics, and more coming, so I do not recommend setting this as a goal.

I would instead focus on color systems routinely used for interchange, namely:

- sRGB
- ITU Rec. 709 (minor differences from sRGB)
- ITU Rec 2020 (Wide Gamut content)
- ITU Rec 2100 (for HDR and Wide Gamut content) [same color primaries and white point as ITU Rec 2020]

> The "p3" color space

There is no such thing as a P3 color space (see https://lists.w3.org/Archives/Public/www-style/2016Sep/0092.html)

More importantly perhaps, the P3 color gamut essentially fits within the ITU Rec 2020/2100 gamut.

### LINKED ISSUE #16: Absolute vs relative luminance
> float values outside of [0,1] range can be used to represent colors outside of the chosen color gamut.

Care will be needed here since at least one popular HDR color systems, i.e. ITU Rec 2100 PQ, maps to absolute linear light values, which range from 0 to 10,000.

In this specific case, it will be necessary to specify that the canvas value are for `normalized linear colour values` to use the term from ITU Rec 2100 PQ.

palemieux:
Closes #23 
Closes #15 
Closes #16

### DIFF
--- a/CanvasColorSpaceProposal.md
+++ b/CanvasColorSpaceProposal.md
@@ -19,7 +19,8 @@ Non-goals of this proposal are:
 * Ultra-wide color gamut color space support (e.g, Rec2020) and high dynamic range support.
   * This topic is deferred to a subsequent proposal.
 * High bit depth 2D Canvas and ImageData support.
-  * For 2D Canvas and ImageData, this topic will be addressed by UWCG and HDR proposal, because such color spaces require more than the default of 8 bits per pixel.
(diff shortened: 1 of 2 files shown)


palemieux:
@ccameron-chromium @svgeesus I would appreciate your review.

svgeesus (approved):
LGTM

ccameron-chromium (approved):
lgtm

PR #65, opened 2022-02-24
Fix stray commas and use more modern JS

tomayac:


kdashg (approved):
I do think this is nicer.

kenrussell (approved):
LGTM too.

